PR #34, opened 2013-06-12
Added required property

jbrooksuk:
This will require styling on the developers part, however it's easy to check whether the Selectized element has `full` or not.


baptistedonaux:
@jbrooksuk Thank you for your solution !

@brianreavis Add native support for required option will be interesting.

PR #93, opened 2013-08-29
set ID for selectize control based on the ID of the original input

papermoon1978:
Not sure if you need this. In my case I selectize some hidden (display:none) input elements. And having an ID for the selectize wrapper it's much easier to show/hide the control on demand.


alltonp:
Please please please can we have this. Right now automated testing is very painful/brittle and for me that was one of the primary reasons for moving away from select2.

PR #185, opened 2013-11-05
Adding support for focus and blur events

JedWatson:
Hi,

First of all, thanks for your great work on this!

I added support for focus and blur events on the control. It watches the previous state of the .isFocused boolean to ensure the event is only fired when the focus state changes, rather than when the internal focus event is called.

I've got a use case where I want to clear the value on focus, then reset it on blur unless it has been changed, which makes selectize work more like the URL bar in a browser.

But I expect it will be useful for other cases as well, including more sophisticated interaction with other UI controls, etc.

Cheers,
Jed.

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -33,6 +33,16 @@ selectize.off('event_name', handler);
 		<td valign="top">Invoked when the value of the control changes.</td>
 	</tr>
 	<tr>
+		<td valign="top"><code>"focus"</code></td>
+		<td valign="top"></td>
+		<td valign="top">Invoked when the control gains focus.</td>
+	</tr>
+	<tr>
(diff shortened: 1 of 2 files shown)


jacquescrocker:
+1 I'm in need of these events as well

rzhou186:
+1 Needing focus and blur as well.

brianreavis:
Holy hell that took a long time to get to. Sorry it took so long, all. Thanks for the great work @JedWatson and @scarfacedeb!

super-cache-money:
Can we build dists from this?
I'd love to be able to get the standalone...

brianreavis:
I've merged a bunch of commits and am making some changes – a release will be out soon

PR #195, opened 2013-11-19
Create a plugin to restore basic label[for=“fieldid”] click / focus func...

ghedwards:
...tionality


krnlde:
Why is this still open?

PR #295, opened 2014-02-19
If array is passed into addOption then respect the insert order

### LINKED ISSUE #218: Disable automatic sort
Hi, I'm receiving a pre-ordered list (JSON) but when I loaded into the selectize control, it re-orders the list by the Id and I don't want this. How I can let selectize use the order of my JSON?

This is the JSFiddle example:
http://jsfiddle.net/LYYab/

Thanks.

erlingur:
Here is a jsFiddle demonstrating the problem: http://jsfiddle.net/8H62B/6/

I'm not a great Javascript programmer so if I did something wrong please feel free to point it out.

When addOption is given an array I would like it if it would respect the order of the array. We already have an $order property on the option so I just fill that in with the appropriate value.

I wasn't sure what files I should change so I just changed everything except the minimized versions. This should hopefully fix #218. Any thoughts?

### DIFF
--- a/dist/js/selectize.js
+++ b/dist/js/selectize.js
@@ -1492,14 +1492,18 @@
 		 *
 		 * @param {object} data
 		 */
-		addOption: function(data) {
+		addOption: function(data, order) {
 			var i, n, optgroup, value, self = this;
 	
 			if ($.isArray(data)) {
 				for (i = 0, n = data.length; i < n; i++) {
-					self.addOption(data[i]);
+					self.addOption(data[i], i);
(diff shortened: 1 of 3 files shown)


brianreavis:
Thanks for the pull request! I'm going to take a different approach, but agree with you – this is something that needs solved. See https://github.com/brianreavis/selectize.js/issues/640#issuecomment-71788203

erlingur:
Hey @brianreavis, awesome, thanks! Good to hear :)

PR #302, opened 2014-02-21
Added 'tabindex' config option

neilb:
If specified, then a tabindex is added to the textbox. A tabindex only makes sense in some situations,
but in those situations, it can be right handy.

Updated usage doc as well.

Cheers,
Neil

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -115,6 +115,12 @@ $(function() {
 		<td valign="top"><code>false</code></td>
 	</tr>
 	<tr>
+		<td valign="top"><code>tabindex</code></td>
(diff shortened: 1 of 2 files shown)


miguelcobain:
:+1:

joallard:
What if we copied the tabindex from the original input instead of adding an option? We're suffering from optionitis a little.

miguelcobain:
I agree.

We always can copy if present on the original element, but fallback to an option if not.

joallard:
Thanks for the PR, while waiting for a response/reaction, I'll close this to facilitate the management of other PRs. Feel free to re-open or re-submit with changes.

PR #324, opened 2014-03-10
Closes #129, allows duplicate values

### LINKED ISSUE #129: Allow duplicate values in an input
Is it possible to have an `input` element which allows duplicate values? The reason is, I'm trying to build a form which allows people to enter ages of children, which can of course, be the same; twins, triplets etc.

jbrooksuk:
I broke down and had to add this in! For duplicates to work, you must have the `mode` set to `multi`, otherwise it wouldn't make sense.

Documentation changes to follow.

I've included the compiled files too. Since I only added the `duplicates` option, I don't know why there are so many other changes? If you can figure this out, I can change the commits.


jbrooksuk:
Closing due to conflicts and no time to make it work.

PR #371, opened 2014-04-10
Add selectOnTab config option

ryanwersal:
Adds a new config option called "selectOnTab" to make the functionality added in #258 optional since it is undesirable behavior in the event that there are many, potentially not required, Selectize controls in a single form (as just tabbing through them will cause the first option to be selected in each). 

I ended up defaulting it to false as this was the "default" behavior prior to 0.9.0.


deus-ex:
Glad I saw this now, wandering why when I press tab it tabs to the next input element and doesn't select the selectize input

Issue #29, opened 2013-06-08
Allow dropdown to open in the "up" direction (and "auto")


brianreavis:
``` js
dropdownDirection: "auto" | "up" | "down"
```

(handle within `positionDropdown` method)

lancedikson:
+1

gwezerek:
Tambien, would be a boon.

booleanbetrayal:
+1 ... would be great to have the `auto` feature!

mrova:
+1!

Jackpohare:
+1

impworks:
I have created a pull request implementing this feature. If anyone is still interested, you can give it a try.

atoftegaard:
@impworks I see several problems with your pull request.
- `dropdownDirection` takes the default value from settings.mode, which is either `single` or `multi` - never auto. You should use a different property for specifying `auto | up | down`.
- `dropdownHeight` is always 0 because you get the value while the dropdown is still hidden. You can only get element sizes when they are visible.
- You do not maintain the borders around the dropdown. Depending on the direction, either top or bottom border should be visible.

After fixing this, I got the results I was looking for.

impworks:
@atoftegaard, thank you for your reply. I am sorry for rushing the pull request without having tested it thoroughly myself. I will try to update it and fix the problems.

meggers:
@atoftegaard, can you make a PR with your changes?

kikonen:
Chances to get this merged?

atoftegaard:
I will see if can get to it friday. I have been very busy since I last looked this.

atoftegaard:
Sorry but I cannot set aside the time to figure out how to push these 4 lines of code, as I haven't made a pull request before (or a modification to a pull request). I'm sure someone else can do it from my comment.

pederjohnsen:
Any reason why this is not yet implemented (seems like a pretty important feature) and are there any working plugins that does this atm?

lmi11:
I really need this feature. How to use it in official version of selectize.js?

kylej-89:
I have something working for you all that doesn't require a pull request. Just requires some additional logic when implementing your settings. I wrote this pretty quick, seems to work ok but I am bound to have missed something here.

    var exceededWindowHeight = false;

    settings.onDropdownOpen = function ($dropdown) {
        var $control = this.$control;
        var offset = this.settings.dropdownParent === 'body' ? $control.offset() : $control.position();                    

[...]

PR #437, opened 2014-05-23
dropdownDirection

streamside:
Thanks for a great plugin! I have been implementing it across our product over the last weeks.

I encountered some problems with dropdowns being clipped when appearing close to the bottom of the screen. I found your comment in one of the issues to implement `dropdownDirection` (#29).

Also added support to reduce height of dropdown in case it can't fit in the window (both up and down).


dule:
@streamside Thanks for the li'l plugin--works great--but there seems to be a bit of an issue when removing a selection. Because `positionDropdown` is called prior to `refreshOptions` during [deleteSelection](https://github.com/streamside/selectize.js/blob/dropdowndirection/src/selectize.js#L1915L1916), the dropdown list does not contain the item that will be re-added to the dropdown list when `positionDropdown` is called, so the dropdown list does not get positioned correctly.

As a bit of a hacky solution, I had to override `deleteSelection` and forced another `positionDropdown` in order for it to position correctly:

``` JavaScript
    this.deleteSelection = (function() {
        var original = self.deleteSelection;
        return function() {
            var deleted = original.apply(this, arguments);
            if (deleted) {
                // Force another re-position after options have refreshed
                self.positionDropdown();
            }
            return deleted;
        };
    })();
```

EDIT: Scratch overriding `deleteSelection`; seems like that's not sufficient, as there are other cases where the dropdown list is updated, and not re-positioned. So took the hack one step further, and called `positionDropdown` whenever `refreshOptions` occurs :/.

``` JavaScript
    /**
     * Force a re-position when dropdown list is updated
     */
    this.refreshOptions = (function() {
        var original = self.refreshOptions;
        return function() {
            original.apply(this, arguments);
            self.positionDropdown();
        };
    })();
```

tjperovich:
@streamside @dule Implemented this in under a minute and it works great so far. Thanks a ton!

rockResolve:
Thanks a lot for this @streamside

The plugin should use its own options.direction, not selectize settings.dropdownDirection
It should also default direction to 'auto' like the original PR.
Code below also falls back to old settings.dropdownDirection if options.direction not set.

Add just after var self = this

```
    options = $.extend({
        direction: self.settings.dropdownDirection || 'auto'
    }, options);
```

And at the start of getDropdownDirection(positions) {

```
        //var direction = self.settings.dropdownDirection;
        var direction = options.direction;
```

johnknoop:
How about a `dropdownDirection: 'auto'` that listens to scroll/screen resize and changes the direction accordingly, like a native dropdown?

jaridmargolin:
@joallard @streamside @ftalves @brianreavis We all seem to want to get this feature implemented and merged. I am happy to complete the remaining work if we can outline exactly what needs to be done.

So a few questions?
1. What is blocking this PR from being merged? I see that no test has been written. I would be happy to try to wire this up (I will have to look at the current test implementation in a little more detail to make sure it won't be a problem).
2. @streamside mentioned a few concerns [here](https://github.com/selectize/selectize.js/pull/437#issuecomment-45409866). Do we consider these essential to the PR. IMO caching the direction would be more desirable than having the input jump from top to bottom. I would be happy to do the work if everyone can agree on this. As for your second remark, specifying the required offset seems like more of an enhancement than a requirement.

---

If we cannot seem to get the functionality correct on `auto`, we could implement `top` and `bottom` with `bottom` being the default. While this doesn't solve all use cases, it definitely introduces new desired functionality.

So essentially first pass would be supporting `top`.
Second pass could be supporting `auto`.

Thoughts?

---

**I want to note that I am willing to give this task full attention until it is merged.**

dashtaki:
I wrote(according to my need) some code that use it in [ui-grid](https://github.com/angular-ui/ui-grid) to show selectize in up direction when it does't have space to open and reverse , it working in ui-gird and any parent element ... 
if (typeof Selectize != "undefined") {

```
Selectize.prototype.positionDropdown = function () {
    var $control = this.$control;
    var $window = $(window);

    this.$dropdown.css({"position": "fixed"});

    var control_height = $control.outerHeight(false);
    var control_space_above = $control.offset().top - $window.scrollTop();
    var control_space_below = $window.height() - control_space_above - control_height;
    var dropdown_height = this.$dropdown.outerHeight(false);

    var up = ( dropdown_height > control_space_below) ? 'true' : false;

    var offset = this.settings.dropdownParent === 'body' ? $control.offset() : $control.position();
    offset.top += $control.outerHeight(true);

    if (up) {

        offset.top = control_space_above - dropdown_height;
        offset.left = $control.offset().left;

        this.$dropdown.css({
            "border-top-color": "hsl(0, 0%, 72%)",
            "border-top-width": "1px",
            "border-top-style": "solid",
            "border-bottom-width": "0",
            width: $control.outerWidth(),
            position: "fixed",
            top: offset.top,
            left: offset.left,
            display: "block !important"
        });
    }
    else {
        this.$dropdown.css({
            width: $control.outerWidth(),
            top: offset.top,
            left: offset.left,
            position: "absolute",
            "border-bottom-color": "hsl(0, 0%, 72%)",
            "border-bottom-width": "1px",
            "border-bottom-style": "solid",
            "border-top-width": "0"
        });
    }
};
```

}

joallard:
@jaridmargolin:

> So essentially first pass would be supporting top.
> Second pass could be supporting auto.

Totally agree, let's take the iterative approach here. 

> I want to note that I am willing to give this task full attention until it is merged.

My main concerns is that this doesn't break anything, doesn't get broken in the future (has tests), works on supported browsers, and is reasonable in complexity, coding and solution design.

If you can provide me with reassurances here and add sensible tests, ideally some peer review, I'll be glad to merge this, it's a functionality that people are eager to have.

imapollo:
Is there any progress on this issue or any workaround?

dashtaki:
https://github.com/dashtaki/selectize.js/pull/1

PR #449, opened 2014-05-29
Add clearInput flag to clearOptions()

ttback:
Only clear user input when clearInput is true or undefined.

The undefined check is to make sure this change doesn't break any existing code.

### DIFF
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -1186,7 +1186,7 @@ $.extend(Selectize.prototype, {
 	/**
 	 * Clears all options.
 	 */
-	clearOptions: function() {
+	clearOptions: function(clearInput) {
 		var self = this;
 
 		self.loadedSearches = {};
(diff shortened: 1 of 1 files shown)


u01jmg3:
@brianreavis: can you give a recommendation on each new search -- after selecting an item -- how to empty the cache of current results? As with those above, each new item added results in the previous results becoming invalid. I therefore want to start from scratch after the item add, and rely on what my remote data source returns.

Would it be better to use `updateOption()` / `removeOption()` instead? Would they keep any items already selected? The issue is preserving items already selected but clearing previous results from the dropdown.

I appreciate removing `self.clear()` from `clearOptions()` is not the way to do this.

u01jmg3:
Solved using:

``` js
var $selectize = $('.selectize').selectize({
    onItemAdd: function(){
        var options = $selectize.options;
        for(option in options){
            if($selectize.items.indexOf(option) == -1){
                $selectize.removeOption(option);
            }
        }
    }
});

$selectize = $selectize[0].selectize;    
```

PR #477, opened 2014-06-16
Clear the selected item when the user types any alphanumeric character

mikeanden:
When the dropdown is configured for single selection and an item has been selected,
this will automatically clear the selected item when the user types to search for a
new item. With the current behavior, the user has to press back space to clear the selected
item. The UI seems to be unresponsive if the user wants to change the
selection unless they hit backspace

### DIFF
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -485,7 +485,10 @@ $.extend(Selectize.prototype, {
 		}
 
 		if ((self.isFull() || self.isInputHidden) && !(IS_MAC ? e.metaKey : e.ctrlKey)) {
-			e.preventDefault();
+			if (self.isOpen && self.$activeOption) 
+				self.setValue('');
+			else
+				e.preventDefault();
 			return;
 		}
 	},


dule:
Changeset in plugin form:

``` JavaScript
Selectize.define('clear_on_type', function(options) {
    var self = this;
    var IS_MAC = /Mac/.test(navigator.userAgent);
    var KEY_BACKSPACE = 8;

    this.onKeyDown = (function() {
        var original = self.onKeyDown;
        return function(e) {
            var originalCode = e.keyCode;
            if ((self.isFull() || self.isInputHidden) && !(IS_MAC ? e.metaKey : e.ctrlKey)) {
                if (self.isOpen && self.$activeOption) {
                    e.keyCode = KEY_BACKSPACE;
                    self.deleteSelection(e);
                    e.keyCode = originalCode;
                }
            }
            return original.apply(this, arguments);
        };
    })();
});
```

dule:
@feliperaul I'm actually not a contributor of selectize, merely a user of it providing a workaround. So thanks should not be directed at me :). Also means I can't answer your second point :man_shrugging: .

feliperaul:
@dule I think my phrasing was ambiguous :) I'm thanking you for the provided solution, works perfectly! Also, I know that you don't have commit access, the question 'why isn't this merged' wasn't directed at you :) Thanks one more time!

dule:
Ah gotcha, plugin as in the changeset plugin format. I was interpreting plugin as the selectize plugin. Whoops. Well... Your welcome :)

TakesTheBiscuit:
@dule @feliperaul Could you lend me a quick hand with the changeset plugin above?
I've included it before all of my JS and then i'm instantiating my selectize like so:
```
 $select_country = $('#select-country').selectize({ allowEmptyOption: true, clear_on_type: true });
```
Am i getting this right? - not sure how to instruct selectize to use the `clear_on_type` plugin

- doh, rtfm!
`    plugins: ['restore_on_backspace'],`
got it!
scroll down: http://selectize.github.io/selectize.js/
look for 'demo' tab, click 'plugins' tab, there it is, hiding away in plain sight 
<3 thanks 

i'm now using
```
$select_country = $('#select-country').selectize({plugins:['clear_on_type'], allowEmptyOption: true });
```
perfect!

PR #485, opened 2014-06-20
Disabled options

daimor:
options marked as disabled, in html, or in JSON, shows as disable and unselectable


alperensaylar:
@designermonkey how did you solve render issue?

designermonkey:
Unfortunately, I no longer remember :(

PR #542, opened 2014-08-18
Some css discrepancies between selectize and bootstrap 3.2.0

sod:
I fixed some issues with the bootstrap3 styles:
- bootstrap3 input transition
- support for .has-success, .has-warning and .has-error
- add right padding on .has-feedback
- fix height difference between input and selectize

See http://plnkr.co/edit/rci0eIvARvY3nEMQ40uq?p=preview

Tested in:
- Chrome 36
- Firefox 31
- Internet Explorer 10
- Internet Explorer 9

### DIFF
--- a/src/less/selectize.bootstrap3.less
+++ b/src/less/selectize.bootstrap3.less
@@ -97,6 +97,7 @@
 }
 
 .selectize-input {
+	overflow: inherit;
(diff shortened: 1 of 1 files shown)


sod:
No feedback in over a year. I guess that means no.

PR #557, opened 2014-09-04
Add option for readonly selectizes

mfouquet:
Sometimes you just want a consistently skinned <select> replacement. Helps when you don't want the keyboard popping up on mobile devices, too.


dcramble:
It doesn't appear that a "readonly" attribute transfers from the base select/input element to the selectize input element. It seems this code could solve that problem in a more manageable way.

Issue #593, opened 2014-10-11
clearOptions should not clear items


ghost:
The clearOptions() method results in the clear() method also being called. Given the logical distinction between "options" and "items" this seems incorrect. The behavior also prevents (unless I'm missing another approach) the use case of providing a new set of options for each item, as follows:
    $('#select-state').selectize({
        load: function(query, callback){ 
            this.clearOptions();
            //fetch the new options
        }
    });

dresende:
This is nonsense why it's still open. You should just remove `.clear()` at the end of the function and it just works as expected.

dipbazz:
Has this issue been fixed? if not then how can we implement clearOptions with a hacky way so that our options will be cleared without being cleared the selected input option? If Somebody can help then it would be really appreciated.

jicihome:
According to this changelog:
Fixed bug making clearOptions function. Now it doesn't remove already selected options.

(thanks @caseymct - #1079)

But it doesn't seem to work.
Also, silent is implemented and doesn't see to work too.

jicihome:
I can confirm for silent. Not working for add item

KES777:
This is wrong proposition. `clearOptions` must call `clear` otherwise data will be inconsistent. 

For example you have list 1,2, user selected 7. Is this expected? No, because there is no 7 on the list.  
Same with clearOptions. If options are cleared, you can choose nothing, thus `selected list` also must be cleared.

suchoss:
Imho, this is bad design - method is called `clearOptions` and that is exactly what is expected from the method to do. 
It shouldn't delete any already selected items. 

For example in my case - I need to delete all options all the time, because I am calling backend with every pressed letter on keyboard. Then user selects what they need and I close options and delete them - ready for next query. When they select everything they need, they submit results with button. In version `0.12.6` this worked quite well.

jthakilla9:
> This is nonsense why it's still open. You should just remove `.clear()` at the end of the function and it just works as expected.

Its true what he said, remove the clear and use it like this example:  

it works like butter 👍 

// Functie om gebruikers op te halen en dropdown te vullen met behulp van Selectize
function loadUsers(selectedEmails = []) {
  Fliplet.DataSources.connect(usersDataSourceId)
    .then(connection => connection.find())
    .then(users => {
      if (!users.length) {
        console.warn('Geen gebruikers gevonden.');
        return;
      }

      const emailSelect = document.getElementById("emailSelect");
      
[...]

PR #595, opened 2014-10-14
Fix to issue #593.

ghost:
Proposed fix to issue #593.


starg0ddess:
When I apply this fix, it creates an error that it can no longer find my searchField name. I've linked up a list of objects as my initial data source. When I select a first option, no problem, but when I select a second option (based on a new user typed search) the error is generated.

Pictor13:
@starg0ddess is it still the case?
Could you give a working snippet of the described bug?

Issue #406, opened 2014-05-12
Clicking on another input when selectize has focus does not fire 'focus' event on input (Internet Explorer)


benmj:
Consider a situation where you have a selectize control and another input ("Input A"). The second input has a handler registered to the 'focus' event. In Internet Explorer, if the user clicks into the selectize control and then clicks on "Input A" the focus event is not fired.

The problem seems to be something in selectize's `onBlur` function, but I haven't been able to pin-point it. Also, the issue is not observed when using the "tab" key to change focus between elements.

I've observed this in IE 11 and simulated versions of IE10 and IE9.

You can see a plnkr of the issue here: http://embed.plnkr.co/E0jodB0liZKCQi7iu2P6/preview -- you'll need to have the console open

PR #599, opened 2014-10-17
Ensure correct IE focus when the user clicks outside the focused control. #406

perrej:
Clicking on another input element when selectize has focus causes it to focus on the body in IE. By calling focus on the target after blurring, the focus event is fired as expected on IE.


AndersDJohnson:
@brianreavis Looks like you merged this, then later reverted in 6f409616623b940d41dfdc8576ccd8d7b8346cc5? Can you explain? Related to #406.

PR #608, opened 2014-10-28
Added LESS file for Free Bootswatch's Flatly theme.

thiago-negri:
I don't know if you would like to maintain aditional LESS files for Bootswatch themes.
But I've tweaked the `selectize.bootstrap3.less` file a bit so it would look better using Flatly theme.
The lines I've edited are marked with comments.

Essentialy:
- `@selectize-border` changed to `2px`
- Added transition to `.selectize-input`
- Removed box-shadow of `.selectize-input .focus`
- Changed `.selectize-dropdown .active` a bit.


brianreavis:
Thanks for the contribution, although I feel Selectize is already getting to be too much of a behemoth to maintain. I feel like people [searching](https://www.google.com/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=selectize%20bootstrap%20flatly) around could find this interesting though

PR #629, opened 2014-11-25
Allow jumping to next field with Tab if control is full

toupeira:
This fixes the behaviour introduced by https://github.com/brianreavis/selectize.js/pull/352
Synthesizing keyboard events is really tricky so I didn't bother writing tests for this.

Other related issues:
- https://github.com/brianreavis/selectize.js/issues/152
- https://github.com/brianreavis/selectize.js/issues/350
- https://github.com/brianreavis/selectize.js/issues/396


toupeira:
Thanks!

PR #639, opened 2014-12-04
Remove rendercache when removing all options

jperelli:
Regarding this https://github.com/brianreavis/selectize.js/issues/317


brianreavis:
This was fixed by https://github.com/brianreavis/selectize.js/commit/dbca6b0b5e5212c69472693053935c9492a14122

PR #657, opened 2014-12-28
Use <button> elements instead of <a href="javascript:void(0)"> in remove_button plugin

rossshannon:
Using `<a href="javascript:void(0)">x</a>` to add an interactive button that will be triggered by JavaScript is a little bit 90s. The `<button>` element is a more appropriate click target in the remove_button plugin, and works identically. I have tweaked some of the styling so that the button doesn’t inherit any built-in browser styles.

I’ve been using this tweak in a project of mine since May and have tested in all major browsers back to IE9.

screen shot 2014-12-28 at 20 27 36


brianreavis:
Out of curiosity, why is the height hard-coded to 22px?

PR #666, opened 2015-01-13
Usage.md correction

parsonsmatt:
I attempted to use selectize.js with a `create: function(input, callback) {}` pattern, as suggested by the docs. This wouldn't work, as the function never returns anything.. After some light debugging, I realized that the `callback` function wasn't being used or defined. I tried setting up the function given in the example, and it worked. I checked some of the examples with this definition, and they all followed the given format.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -37,7 +37,7 @@ $(function() {
 	<tr>
 		<td valign="top"><code>create</code></td>
 		<td valign="top">
-			Allows the user to create a new items that aren't in the list of options. This option can be any of the following: "true", "false" (disabled), or a function that accepts two arguments: "input" and "callback". The callback should be invoked with the final data for the option.</td>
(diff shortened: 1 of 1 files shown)


brianreavis:
Hm, did you dry invoking the callback with your intended data? Here's where it's being passed a function: https://github.com/brianreavis/selectize.js/blob/master/src/selectize.js#L1438

``` js
// these do the same thing:
create: function(input, callback) { return {tag: input}; }
create: function(input, callback) { callback({tag: input}); }
```

Selectize expects that you either return an object immediately (synchronous), or invoke the callback with an object (async).

parsonsmatt:
That works as described -- thanks! I will update the PR to reflect that behavior of the function.

I ran into some initial confusion where the Selectize was silently failing to add things to the search list due to improper setup -- eg, `create: true` works fine with an otherwise bare configuration object, ~~but `create: function(){}` only works if the `valueField` and `labelField` variables are set. I'll add notes about this as well.~~ wups I didn't read their default values further down the page! I need to stop coding so late. 

Since the function has some expectation about it's input, would it be a good idea to throw exceptions or otherwise notify that improper input was used?

PR #680, opened 2015-01-21
All event handlers will be skipped when mousedown event is fired because...

christiaan:
... of return false

When selecting a already selected item and then focussing the
text input part of the select does not open the dropdown

Only happens on inital load. 
To test go here http://brianreavis.github.io/selectize.js/ and go to the Max Items demo. Click Wyoming to select it and then directly after click inside the textfield to the right of it. The dropdown does not open. You first have to unfocus the selectize and click in it again to open the dropdown.


brianreavis:
What browser are you using? I can't reproduce this in Chrome

christiaan:
I've just reproduced it using these steps on Linux with Firefox or Chrome. And on Windows with Chrome, Firefox and Internet Explorer.

To recap what I do exactly.
I open http://brianreavis.github.io/selectize.js/
Scroll down till I see the "States:" select that contains California and Wyoming.
I click once on the word Wyoming.
I click on the white space on the right next to Wyoming inside the text field.

The dropdown does not open.

When you do the same steps but don't click Wyoming once it works perfectly.

ibcooley:
I am also experiencing this issue.  It is quite inconvenient for our users.

squadette:
If you're interested, I have this bugfix tested on a wide variety of users and it works as advertised.

PR #682, opened 2015-01-21
Configurable `loading` class name

duncanbeevers:
The `$wrapper` element is given a `'loading'` class name while asynchronous load requests are outstanding.

This pull request provides a configuration point for overriding the default value of `'loading'` during this operation.


brianreavis:
Thanks!

PR #687, opened 2015-01-23
suppressAddDuplicate option

duncanbeevers:
Hides "Add..." control when the entry already exists in the known options


brianreavis:
This is what the [createFilter](https://github.com/brianreavis/selectize.js/blob/master/docs/usage.md) setting is for – it can't always be assumed that an option's value will by the input text.

``` js
{ createFilter: function(input) { return this.items.indexOf(input) > -1; } }
```

duncanbeevers:
Thanks for the clarification.

PR #698, opened 2015-01-30
passing silent value to clear method to ensure events are not raised

bubb-rubb:
New silent parameter was not being passed into the clear method in setValue and addValue


brianreavis:
Thanks for the catch!

PR #708, opened 2015-02-06
Use @input-height-base instead of auto

stephanvierkant:
.form.control.selectize-control should use @input-height-base, like other form fields do.


ShaggyDude:
+1 Yes please, I need to make the default editor height smaller!

PR #709, opened 2015-02-06
Dropdown shadow in less variable

stephanvierkant:
The box shadow is hardcode, so adjusting/removing the shadow isn't possible. Therefore, the dropdown shadow should be in a less variable.


brianreavis:
Can you add the missing semicolon so that it builds? Then I'd be happy to merge this

stephanvierkant:
I'm sorry! I just added the semicolon. Thanks for merging!

Issue #713, opened 2015-02-11
Argument error in selectize.less causing compiler to abort


urban:
The current version of selectize.js has a bug in the Less file. At approximately line 267 of serialize.less.  You need to add a second set of parenthese to this line. 

```
// change from 
margin-top: round(-1 * @selectize-arrow-size / 2);
// to
margin-top: round((-1 * @selectize-arrow-size / 2));
```

PR #714, opened 2015-02-11
Fixed argument error in selectize.less

urban:
This fixes issue #713


brianreavis:
Thanks. Genuinely curious... do you know why this is a problem? It is it a buggy release of less? This can't be expected behavior, can it?

urban:
I really don't know. I created this [gist](https://gist.github.com/urban/18bcc973d69887068c5f) to try and isolate it but I'm not getting the same error any more. In my opinion, this is probably an issue with less, however I wouldn't know for sure without more digging.

PR #717, opened 2015-02-17
Fixed scoping issue which prevented multiple instances from having similar options

IanCaunce:
If there are multiple instances of selectize on a page and each instance shares an option with the same value upon initialization, only the first instance will have their option pre-populated. This fix changes the scope of the optionsMap so that values are not shared between selectize instances.


brianreavis:
@IanCaunce Thanks!

PR #720, opened 2015-02-20
Added searchOnKeypress setting, fixed Firefox and IE tabbing behavior

### LINKED ISSUE #396: Bug: Tab focus behavior in Firefox
When a selectized select control is used in a form, the tab behavior is important.  Currently in Firefox, if the option list is long enough to need to scroll, then tabbing out of the selectize input box does not focus the next control.  The document.activeElement indicates that the body element is active.  Tabbing a second time will either focus the next control or refocus the selectize input box depending on whether there is a selection.  This is inconsistent with other browsers and awkward for keyboard navigation.

Demo: http://jsbin.com/xitom/1/
Affects version: 0.9.1
Broken in: Firefox 28, 29
Works in: Chrome, IE
Workaround: Use explicit tabIndex

dcwatson:
Should fix #396 (there are actually 2 distinct issues in there - the original Firefox issue, and an IE tabbing issue), and provides a way to start searching when the user types, ala #477 but with a setting to turn it off.


dcwatson:
I just realized I made the changes to the dist file. I'll submit a new PR to the source.

PR #723, opened 2015-02-23
Fix hash_key checks so allowEmptyOption works again

dcwatson:
A hash_key of `''` is a legitimate value if `allowEmptyOption` is set to true. Change the checks from `!key` to `key === null`.


rainhead:
Works for me. We'd love to get this merged!

davilious:
+1, works for me also, I hope to be merged soon :+1:

PR #726, opened 2015-02-27
Allow reloading of updated options from html

p91paul:
This pull request provides the class Selectize with a method to reload options from the underlying html element. It is useful when combined with other plugins that add <option> tags to the select, like https://github.com/dnasir/jquery-cascading-dropdown


p91paul:
This pull request was never reviewed, and I'm not a selectize user anymore. So if there is no interest it can be closed.

PR #741, opened 2015-03-18
Disable feature

Teraglehn:
Add disabled option and option group support in selectize. With updated documentation and API.


vylan:
@Teraglehn  I've tried your code, I found that when I disabled an option, it still can be selected. Is there any way to make the option unselectable?

Teraglehn:
Can you expain how you select it ? (By using UI or API ? I guess API, i'm not 100% sure of having disabled all entries)

Teraglehn:
So, hum, now it's done :) Sorry for waiting ;)

PR #758, opened 2015-04-05
Do not alter the form control height

kadimi:
Hi,

Please refer to the example and comments here: http://codepen.io/kadimi/full/KwLVap


kadimi:
Please note, this is the external workaround I use for now:

```
/**
 * Fix selectize elements height
 */
.form-control.selectize-control {
    height: @input-height-base;
}
```

**Edit**
But this prevents expanding the field if needed.

PR #762, opened 2015-04-14
Don't lose the dropdown's focus when the user clicks on an optgroup heading

bradchristensen:
When you have lots of option groups in a list and a small number of options in each group, the user may accidentally click on an optgroup heading, which blurs focus from selectize and closes the dropdown. This seems to be unintentional behaviour given that clicking on any other empty area of the dropdown (not including options) will not cause focus to be lost.

This fix checks to see if the class name of the element being clicked on is that of an optgroup heading and if so, prevents focus from being lost.


u01jmg3:
:+1:

johnknoop:
Any chance of merging this as-is, or at least offering an option 'closeOnOptgroupClick' or something?

Issue #742, opened 2015-03-18
Problems... selecting, in dropdowns.


nkoren:
Generally loving Selectize, but having some subtle-yet-important UX problems with its core functionality. Specifically, the auto-suggestion is too sensitive to some things yet too insensitive to others, and I can't find any way to override its behaviour.

What I want to do is the following:
- When the user types, the top-scoring option is highlighted.
- If they hit tab/enter or click on that option, that option is selected.
- If the top-scoring option is not an exact match to what they've typed, they can click on "Add [what they've typed]..." or arrow down to that and hit tab/enter.
- Simply mousing over the dropdown should have no effect upon its selection-state.

Instead, I'm getting the following:
1. The suggested option is too sticky, and remains suggested even if a higher-scoring option is available. For example, I have a list of options which includes "global trade" and "global warming". If the user starts typing "global", then "global trade" will be the top-scoring item and will be auto-suggested. So far so good. But if the user continues typing "global warming", then "global trade" will remain the suggested option, even though "global warming" is now the top-scoring item. If the user then hits enter/tab, they will select "global trade" despite having just typed an exact match to one of the other options. _This is critically broken functionality!_ How can I fix it?
2. If the user mouses over the dropdown list, then whatever options they've moused over will become the suggested option. This shouldn't happen: I'm seeing users tying something, see that it matches the top option, then accidentally brushing the trackpad before hitting tab/enter. In the process, they mouseover the dropdown and select another item. This leaves them very confused.

Any help will be much appreciated!

nkoren:
The options I'm using, FWIW:

``` text/javascript
{
  plugins: ['continue_editing'], // See selectize issue #181
  openOnFocus: false,
  create: true,
  createOnBlur: true,
  createFilter: canCreate,
  openOnFocus: false,
  maxOptions: 20,
  persist: false,
  highlight: false,
  options: factors,
  valueField: '_id',
  searchField: 'title',
  labelField: 'title',
  sortField: {
    field: '$score',
    direction: 'desc'
  },
  searchConjunction: 'or',
  selectOnTab: true,
  onType: charactersLeft,
  closeAfterSelect: true
}
```

nkoren:
_Bump_. Don't mean to be rude, but problem no. 1, above, is really critically breaking Selectize. It should not be possible for the user to type an exact match to one of your options, hit enter, and have it select _another_ option. Has anybody else encountered this, and can you advise on how to fix it?

gabsoftware:
[...]

PR #770, opened 2015-04-18
2 new settings

gabsoftware:
- Added setting:"ignoreHover"
  When true, this setting will prevent the automatic activation of hovered options.
- Added setting "alwaysActivateFirstMatch":
  When true, will always activate the first option in the dropdown list results

Those settings are useful to always get the same option selected when hitting Enter.

This is in response to issue #742.

### DIFF
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -71,8 +71,7 @@ var Selectize = function($input, settings) {
 	}
 
 	//if specified, do not make options active on mouse hover
-	if (self.settings.ignoreHover)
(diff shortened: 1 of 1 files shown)


nkoren:
+1! These are hugely useful settings which I would LOVE to see merged into the main branch!

antitoxic:
+1 Plugin doesnt work for this (lots of rewrites). Better be in the core.

sergeyk:
:+1:

Issue #647, opened 2014-12-12
Allow typing mode in selectize single selects


luanfonceca:
Hi folks, today i did some improvement in our UX from our selectizes, and the result of that work was a plugin, following i show a gif with the result, if you guys think that this can be merged into the core plugins I am glad to fork e send a pull-request. o/

http://g.recordit.co/hiG2m6gAbF.gif

brianreavis:
@luanfonceca Definitely pull request! This is something I would be interested in making the default behavior actually. Looks great :+1:

stevelacey:
@luanfonceca if you focus the select when it already has a selection, then blur, I assume it retains the selection? If so, this is great, and totally addresses #610

jeepey:
@luanfonceca I'd really appreciate this functionality as well. Could you do the pull request (or share the code)?

alejcas:
This is great but... it is posible to allow the user to go off editting (go back before start typing) when pressing Escape key?

luanfonceca:
Hey @janscas, sorry but i dont have any clue to make that, because i just followed the default behaviour of the _selectize_, I just went to the demo page and press ESC in a `select` to see if there was a bug in my plugin, but my plugin has the same behave of the selectize... 

So i dont have idea to solve that.
Sorry for that o/

dcramble:
I really like this plugin. However, I noticed when the selectize element is setup for an ajax load type configuration, I ave to click on the select field twice to get the data drop down to appear. When I disable this plugin, the issue goes away.

christor:
Is there something preventing this from being merged? This looks like a very nice plugin to include.

risadams:
closing stale issues older than one year.
If this issue was closed in error please message the maintainers.
All issues must include a proper title, description, and examples.

PR #776, opened 2015-04-23
Allow the users to type in selectize single selects.

luanfonceca:
This "PR" refers to #647, I will be glad if you guys have other thoughts on how i did that and sorry for taking so long to do this.

o/

### DIFF
--- a/examples/plugins.html
+++ b/examples/plugins.html
@@ -113,6 +113,35 @@
 				})
 				</script>
 			</div>
+
+			<div class="demo">
+				<h2>Plugin: "typing_mode"</h2>
+				<div class="control-group">
+					<label for="select-language">Languages:</label>
+					<select id="select-language" placeholder="Select a language...">
+						<option value="txt">Text</option>
+						<option value="md">Markdown</option>
+						<option value="html">HTML</option>
+						<option value="php">PHP</option>
+						<option value="python">Python</option>
+						<option value="java">Java</option>
+						<option value="js">JavaScript</option>
+						<option value="ruby">Ruby</option>
+						<option value="vhdl">VHDL</option>
+						<option value="verilog">Verilog</option>
+						<option value="c#">C#</option>
+						<option value="c++">C/C++</option>
+					</select>
+				</div>
+
+				<script>
+				$('#select-language').selectize({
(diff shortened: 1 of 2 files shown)


dcramble:
I really like what you have done hear, I have been use this plugin for a few months now.

One issue I'm running into, is the user clicking on the selector after the value has been set, but not actually changing the value, which seems to trigger a "change" event.

This seems to happen when a user sets the value in the selector, then changes the value in a field to the left of, or any field with a preceding tab order, and tabs out of said field. When they tab out, they land on the already set selctor and this triggers a change event, even though the data is not changed by the user.

If you have events wired to the change handler, they fire and issues may arise.

dcramble:
From my initial testing, it seems the issue is with the `blur` event. If I change the `setValue` call to use the optional `silent` argument, the problem goes away. It appears the `change` event is still firing when the value is actually changed, which is good.

**Original code:**

```
    this.$control_input.on('blur', function() {
        self.setValue(self.typingValue);
    });
```

**With proposed change:**

```
    this.$control_input.on('blur', function() {
        self.setValue(self.typingValue, true);
    });
```

johnnyoshika:
This is a great plugin!  Can you suggest how this plugin can be enhanced so that a user can clear a selection?  Previously the <kbd>Backspace</kbd> key would do that, but that obviously won't work with this plugin.

christor:
Is there something preventing this from being merged? This looks like a very nice plugin to include.

GrumpyCrouton:
This gives me the error:

```
Uncaught RangeError: Maximum call stack size exceeded
    at Selectize.self.trigger (selectize.js?plugins:917)
    at Selectize.self.trigger (selectize.js?plugins:922)
    at Selectize.self.trigger (selectize.js?plugins:922)
    at Selectize.self.trigger (selectize.js?plugins:922)
    at Selectize.self.trigger (selectize.js?plugins:922)
    at Selectize.self.trigger (selectize.js?plugins:922)
    at Selectize.self.trigger (selectize.js?plugins:922)
    at Selectize.self.trigger (selectize.js?plugins:922)
    at Selectize.self.trigger (selectize.js?plugins:922)
    at Selectize.self.trigger (selectize.js?plugins:922)
```

PR #782, opened 2015-04-29
Support for chaining on functions .focus() and .blur()

pilafmon:
Allows convenient usage such as:
`var books = $('#books')[0].selectize.blur().getValue();`


pilafmon:
This PR just adds a return value to a couple of functions that previously returned nothing.  The implementations of the functions themselves remains the same.

It should be very low risk.

risadams (approved):

PR #788, opened 2015-05-08
Make indexOf work with IE8

justinTNT:
I'd rather not use a shim, particularly given it's only for indexOf.


brianreavis:
I like the idea, but can you please not update the dist files? Also, it'd make sense to define a function in [utils.js](https://github.com/brianreavis/selectize.js/blob/master/src/utils.js) vs doing the checks manually everywhere.

joallard:
I find it a little overkill to add lines to the code to support a browser as extinct as IE8, to be honest. What I could get behind is if we had an IE8 monkeypatch just like ES5shim does, but just for the functions we need.

PR #791, opened 2015-05-11
Dynamic opt group implementation

ghost:


ghost:
@brianreavis Would you like me to write integration tests around this?  Also I noticed there is no explicit testing of the utils file, which I could create a new file to start that trend and test my function.  Also side note, do you see the benefit of this change, to dynamically generate the optgroups based on the options?  If not and you would never plan on merging please let me know, and I will maintain a fork instead. Thanks!

PR #792, opened 2015-05-12
Update GitHub Pages documentation for SSL

kevin-brown:
This pull request makes a few changes, I can split these up into separate pull requests if needed.

---
1. This removes the `.DS_Store` file from the repository, which is a binary file that should be in the `~/.gitignore` of anyone on OS X. This also adds `_site` (the compiled directory for Jekyll, which serves the docs) to the `.gitignore`.
2. This switches over most of the remote assets to use SSL, which allows anyone using HTTPS Everywhere to use the documentation and view the examples.
3. This updates the documentation link for [Select2](https://select2.github.io) to point to the correct documentation location.

---

Just a note on point 2 (also mentioned in the commit): Both of the remote API examples do not function over HTTPS because neither of the APIs support SSL. Also, 2/3 of your remote data examples are broken
- **corsheaders.com is no longer operating**, so the states and cities example will never load
- Your Rotten Tomatoes API account has been disabled

Someone might want to fix this so there can be more than just the GitHub example there.


brianreavis:
Thanks, Kevin! I'll look into the other issues.

Issue #110, opened 2013-09-11
Disable text input on a vanilla select


arathael:
Is there any configuration option to explicitly disable text input on the control?

I mean, not to disable the dropdown, but just disable keyboard input (and the caret blinking), something like the readonly attr for the generated text input maybe?

tltowles:
@styledev Thank you!

tltowles:
@efreibe Thank you!

djchar89:
You can disable pointer for the "selectize-control" div. 

example:
$('.selectize-control').css('pointer-events', 'none');

harpreetsb:
Incase this helps

``` javascript
/**
 * triggerSelect
 * @param className
 */

function triggerSelect(className, sort, search, addnew) {
    if (typeof(addnew) == "undefined") {
        addnew = false;
    }

    if (typeof(search) == "undefined") {
        search = false;
    }

    if (sort == true) {
        $(className).selectize({
            create: addnew,
            sortField:'text',
            closeAfterSelect: true
        });
    }
    else{ // get rid of alphabetic ordering
        $(className).selectize({
            create: addnew,
            closeAfterSelect: true
        });
    }

    if (search == false) { // remove input if required
        $(className).siblings('.selectize-control').children('.selectize-input').children('input').attr('disabled', true);
    }
}
```

AndreaCatania:
Have you found a solution? as "styledev" said the input is very very bad on mobile phone.... it's a good library and add 1 option to disable inputbox is very useful for all...

If you add option readOnly where if true the system add the attribute readonly to input element "'<input type="text" autocomplete="off" />'" at line https://github.com/brianreavis/selectize.js/blob/master/src/selectize.js#L124 it's easy.

---

## EDIT 1

### Fast implementation fix

To fast add this features i've changed this line of code:

`$control_input    = $('<input type="text" autocomplete="off" />').appendTo($control).attr('tabindex', $input.is(':disabled') ? '-1' : self.tabIndex);`

WITH

`$control_input    = $('<input type="text" autocomplete="off" />').appendTo($control).attr('readonly',typeof q.attr("data-selectize-readonly")!=='undefined'?true:false).attr('tabindex', $input.is(':disabled') ? '-1' : self.tabIndex);`

now by setting "**data-selectize-readonly**" on the input you will have the input in readonly state.

---

## EDIT 2

~~https://github.com/brianreavis/selectize.js/pull/803~~ #803

zeeberry:
[...]

PR #803, opened 2015-05-21
Implemented "readOnly" property (mobile friendly feature)

AndreaCatania:
Added the property "readOnly", this is used to block user input and prevent keyboard opening on mobile phones.
Ofcourse default value is false and the input is enabled on readOnly true.

This feature was requested some time ago, by other guys #110 today i've implemented this feature because I want to use this (fantastic) library for my project and this feature is essential.


AndreaCatania:
What's wrong?

AndreaCatania:
Why coverage fails? What's wrong in the code? I've tried in local and works fine.

AndreaCatania:
With this coverage i'm forced to write a bad code instead of one "IF" I had to write two "IF"... absolutely useless

drewcovi:
:+1:

samstickland:
+1

kayluhb:
:+1:

filiplikavcan:
+1 This really fixes undesired behaviour on touch devices where keyboard widget shows up when user touches 'selectized select. Thank you.

joseluisq:
:+1:

src/selectize.js
```diff
@@ -126,6 +126,9 @@ $.extend(Selectize.prototype, {
 		$dropdown         = $('<div>').addClass(settings.dropdownClass).addClass(inputMode).hide().appendTo($dropdown_parent);
 		$dropdown_content = $('<div>').addClass(settings.dropdownContentClass).appendTo($dropdown);
 
+		$control.addClass(self.settings.readOnly?'input-readonly':'');
+		$control_input.attr('readonly', self.settings.readOnly);
```
amakhrov, line 130: shouldn't it be `.prop('readonly')` instead of `.attr('readonly')`?

amakhrov:
This PR probably can't be accepted as is.
README:

> When issuing a pull request, please exclude changes in the "dist" folder to avoid merge conflicts.

joallard:
1. Would you please remove changes in dist, as stated?
2. Can you tell me why we're not using attr 'required' instead of adding an option?
3. Can you add tests?

PR #816, opened 2015-06-09
Selectize: sub_menu plugin

edelsbrunner:
Hi all,

I've created a sub_menu plugin. Just created inside of plugins the JS and LESS file.

And updated the plugins.html from examples/ folder.

It work together with the "optgroup", if you define plugins: ["sub_menu"], it will replace the optgroups by the Menu options.

sub_menu
sub_menu2

Best Regards,

Marius Edelsbrunner


edelsbrunner:
Closing in favor of: #818, (I've sent dist/ folder together in this PR)

Issue #752, opened 2015-03-28
Call for Maintainers


brianreavis:
I'm looking for new maintainers of Selectize. In the coming weeks/days, I'll be moving it to an organization account so other people can take care of it. The reasons behind the move (+ some other thoughts):
1. Selectize has been a personal project of mine – it's received no sponsorship and has been a "free time" project. I've been spread thin and it's been increasingly difficult to maintain as my available time has dwindled. I've been heads-down working [Natural Atlas](https://github.com/naturalatlas).
2. _jQuery plugins for UI components have always been a hack._ I'm really stoked by what [React](https://facebook.github.io/react/) is doing in this space. I highly recommend checking out @JedWatson's [**react-select**](https://github.com/JedWatson/react-select) project (even cooler given the fact Jed has contributed to Selectize in the past).
3. Lastly: **open-source burnout**. I work on other OSS projects and jQuery plugins seem to be the most pain afflicting. The burnout stems from a few factors:
   - Everyone has a different idea of how it should work. Different opinions and hard choices are great – but when the stream of people coming into Issues wanting you to drop everything else to make it _their_ way is incessant, it gets old.
   - Entitlement around new features, changes, and support. See bullet 1. 
   - Not enough ingenuity from users (i.e. the Stack Overflow effect). If a style is slightly off, change it for the project. If something minor is broken, fix it and make a pull request that makes obvious what the problem is, how it's solved, and ideally some tests to make sure it doesn't regress. _This is an open source project, not a company's product._
   - Front-end UI has a culture around un-tested code – after all, it's _hard_. Regardless, pull requests with tests are _extremely rare_, and most pull requests come in half-baked. By "half-baked" I mean they come with unintended consequences.
   - The readme is ignored: updating the dist files in a PR has never been the right thing to do.
   - Too many middle-of-the-night support requests.

What I'm mainly looking for in maintainer(s):
- Attention to detail. With the myriad of browser inconsistencies when it comes to focus handling combined with wide array of use patterns, it's easy to naively introduce new issues.

Where selectize needs some love:
- Bugfixes
- Finishing locking down the API.
- The "single" select mode that mimics a normal `<select>` could use some UX improvement. [**react-select**](https://github.com/JedWatson/react-select) is a great example of how it can be improved.

[...]

PR #838, opened 2015-06-26
Add option to disable delete on backspace

vgonda:
Setting disableDelete to true will prevent the user from deleting the previous item when backspace is pressed


Noiwex:
Why you're ignoring pull requests?

brody2consult:
There is a note at the beginning of README.md: "**Selectize needs your help. I'm looking for maintainers!**" which links to the discussion in #752.

PR #859, opened 2015-07-27
Configurable "Add" text translation

atheros:
When create option is set to true, you can input a new item into a selectize field. However the option had a fixed text "Add" as prefix in the dropdown menu.

This pull request makes this text configurable.

### DIFF
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -291,7 +291,7 @@ $.extend(Selectize.prototype, {
 				return '<div class="item">' + escape(data[field_label]) + '</div>';
 			},
 			'option_create': function(data, escape) {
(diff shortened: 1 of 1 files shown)


Alex-Sokolov:
It is configurable now:

``` js
$('#select').selectize({
    render: {
        option_create: function (data, escape) {
            return '<div class="create">Introduce <strong>' + escape(data.input) + '</strong>&hellip;</div>';
        }
    },
    // ... other options ...
});
```

And you can add addtional styling for add row in list

TrishaPatel:
Along with inputed value i want to store static text also(In this case i want to store Introduce too). Can anybody help me?

PR #895, opened 2015-09-01
Using hidden input type instead of text.

victor-torres:
If the 'magic' field will be hidden in any way, why not use 'hidden' input type?

We're currently autofocusing the first text field when a modal is shown in our website. Sometimes the first text element was created by Selectize.js. Then we have to verify this special case before autofocus the form field. If it was a hidden field we wouldn't have this problem.
- All tests have passed.


ghost:
+1

victor-torres:
Selectize fields which require or allow user text input won't work with this patch.

Issue #16, opened 2013-06-05
i18n


marcioaguiar:
Suport for multiple languages.

Text to create new item is hardcoded english.

iainvdw:
+1

klzns:
If anyone is wondering how to change it as @brianreavis suggested, just do something like this:

``` js
$('.my-select').selectize({
  create: true,
  render: {
    option_create: function(data, escape) {
      var addString = 'Adicionar';
      return '<div class="create">' + addString + ' <strong>' + escape(data.input) + '</strong>&hellip;</div>';
    }
  }
});
```

joallard:
As someone who's native tongue is French, I'd be inclined to accept a PR making this simpler. We could add an option allowing you to spectify a string (eg. `addString: "Add %{name}"`) or a function `addString: function(word){ I18n.t("selectize.add", word)}`.

Oxicode:
+1

joallard:
I'll go even further in this, the function would be called with two params, `function(word, key)`, allowing one to do:

`addString: I18n.t`

instead of:

`addString: function(word, key){ I18n.t(word, key) }`

Default: `I18n.t` if exists (or whatever is the JS convention for translation), or `"Add %{name}"`.

or something automagic like that.

joallard:
- [x] Figure out strings that can be translated
- [x] Figure out I18n conventions in JS and sensible defaults
- [x] Begin writing tests
- [ ] Finish writing tests
- [ ] Write the code

joallard:
Here's what I see:

``` js
describe("I18n", function(){
    describe("dictionary syntax", function(){
        var s = Selectize.init({
            dictionary: {
                addThis: "Ajouter %{elem}"
            }
        })

        expect(s.i18n('addThis', {elem: 'ceci'}).to.equal("Ajouter ceci")
    })

    describe("delegate to i18n library", function(){
        var I18n = {
            t: function(key, params) {
                return "Ajouter " + params.elem;
            }
        }

        var s = Selectize.init({
            i18n: I18n.t
        });

        expect(s.i18n('addThis', {elem: 'ceci'})).to.equal("Ajouter ceci")
    })

    describe("default i18n library", function(){
        var I18n = {
            t: function(key, params) {
                return "Ajouter " + params.elem;
            }
        }

        var s = Selectize.init({
            // i18n: default
        });

        expect(s.i18n('addThis', {elem: 'ceci'})).to.equal("Ajouter ceci")
    })
})
```

[...]

PR #902, opened 2015-09-03
Add option for translating labels

nechutny:
Hi,

I've implemented little change that allow translating texts. For example "Add ..." was hardcoded in source. Now is optional config value - for example in Czech: { texts: {add: "Přidat" } }. English is of course still default and don't need any change.


romainsalles:
Depending of the language, the "_Add_" text can be placed before of after the new entity.

I think you should pass a **method**, not a **text** :

``` javascript
{
  'option_create': function(data, escape, texts) {
    return '<div class="create">'+ texts.create(escape(data.input)) +'</div>';
  },
  texts : {
    "create": function(input) {
      return 'Add <strong>' + input + '</strong>&hellip;';
    }
  }
}
```

joallard:
Thank you. We're integrating a more rigid i18n solution in core in #16. Stay tuned for this feature!

PR #910, opened 2015-09-14
Performance upgrade for updateOriginalInput()

deefour:
Previously, on every call to `updateOriginalInput()`, a brand new array of
options for every value in `selectize.items` would be created and the
entire HTML string would replace all children of the raw input source.

When doing something like

``` javascript
selectize.addItems(_.keys(selectize.options));
```

for a few hundred options, this becomes very sluggish.

---

This PR takes a smarter approach, only removing and adding individual
options from the input source as necessary.

The bottleneck is the repeated use of jQuery's `html()` method with large
DOM updates. For ~300 options this was originally spending 30154ms in
`appendChild`. Now it spends **only 330.9ms**.

### DIFF
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -1676,7 +1676,7 @@ $.extend(Selectize.prototype, {
 				return 'option[value="' + value + '"]';
 			});
 
-			if (!fresh.length && !this.$input.attr('multiple')) {
+			if (existing.length - old.length + fresh.length === 0 && !self.$input.attr('multiple')) {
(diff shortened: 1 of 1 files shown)


andrewbaldock:
+1 thank you for this.  Using the drag-and-drop plugin on more than 10 items was unbearably slow, this fixes it.  Nice work.

deefour:
@andrewbaldock I fixed a bug (see above commit) with a blank/default option being added in situations where the select wasn't actually without options.

deefour:
Closing this in favor of #1460

PR #946, opened 2015-11-10
Created click_to_edit plugin with example

krissalvador27:
- Allows users to edit individual selectize items within an input by clicking on them.

example: http://g.recordit.co/q391ZyVBVS.gif

Update:
- Updated to work with selectize called on `<select>` elements and `<inputs>` with maxItems === 1

example: http://g.recordit.co/ZZoImFdKgg.gif


mkimont:
+1 from me as this commit will remove imho biggest UX disadvantage which is lack of editing value which you prior had wroted before.

+2 as you can use 'item' as 'friendly search query'

PR #964, opened 2015-12-05
When preload='focus' and typing the filter, options are incorrectly refreshed with empty filter text.

bkushnir:


SpaceK33z:
The indentation is incorrect here.

joallard:
Yup, I regret to inform OP that Selectize is indented with tabs.

PR #936, opened 2015-10-26
Dropdown direction autodetection.

impworks:
Implements the idea discusses in #29.

PR #966, opened 2015-12-07
implement dropdownDirection option

mlapkin:
Based on #29 and #936


joallard:
Thanks but that sounds like a dupe to #437

Issue #809, opened 2015-06-03
Autogrow buggy with some characters like '@', '.'


apneadiving:
In order to create a new entry (like email), I need periods to be printed (obviously). Everything is fine in chrome but not in firefox unless I add:

```
|| keyCode === 190
```

here: https://github.com/brianreavis/selectize.js/blob/master/dist/js/standalone/selectize.js#L1018

I can't reproduce it in a plnkr so I guess a script of mine could interfere but still, it feels wrong: how come the `printable` is that limited.

I guess its not a bug but does it ring a bell please?

gojohnnygo:
@apneadiving Thanks for find this! I was having the same issue, but I didn't see a PR from you. Do you plan on submitting one? If not, I can do it. Let me know.

joallard:
@gojohnnygo OP says he couldn't reproduce it in a demo, do you experience the bug in a simple use case? I'd like to get to the bottom of this before we just start throwing patches at the code. I'll admit those lines after `printable =` are puzzling.

ellisio:
I'm running into this issue as well on a VueJS application that I'm working on. However, testing a Code Pin using the same code doesn't result in this error in Firefox... The periods work... I feel like it's another library being included.

http://codepen.io/ellisthedev/pen/RrpVBp

I'll keep poking and prodding until I find out what is the cause.

ellisio:
This is not related to just the period key. Any special character `-=[];',./` will fail in Firefox in this condition... Debugging further. I don't think it is related to the `printable` stuff OP mentions though.

ellisio:
From what I can see, this is the issue:

https://github.com/brianreavis/selectize.js/blob/master/dist/js/standalone/selectize.js#L1011

In Chrome, the `keydown` event has an empty value. On `keyup` the value contains the period `.`.
In Firefox, the `keydown` event has an empty value. On `keyup` the value is also empty.

Implementing what the OP recommends _does_ fix it; but it is a hack. This is forcing the `measureString` method to inject the period on the `keyup` event.

I feel like this may be an issue with Firefox itself and not Selectize. But for now, my recommendation is to add the following keycodes:
- `192`: <code>`</code>
- `173`: `-`
- `61`: `=`
- `219`: `[`
- `221`: `]`
- `59`: `;`
- `222`: `'` (Single Quote)
- `188`: `,`
- `190`: `.`
- `191`: `/`

You will also want to add this line after line `1009`:

```
[...]

PR #977, opened 2016-01-07
autoGrow graphical bug fix

johnsenpeder:
There is a graphical bug atm with the autoGrow function where if you type a character not triggering the printable var to be true the input field will not be resized correctly at first, which makes the text "drag" to the left for a quick moment. 
autogrow graphical bug

Also due to using keydown the character string will not have the correct character pressed (depending on keyboard language etc, like when adding @ on a keyboard that requires you to press alt gr + 2 the character would be 2 instead of @).

This is a suggestion that solves this graphical issue and will otherwise act the same way as the original script. (Only tested in FF and Chrome).

If my code is not used directly I hope it will at least help towards improving the original code and fixing this bug.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -271,7 +271,7 @@ var autoGrow = function($input) {
 	var currentWidth = null;
 
 	var update = function(e, options) {
-		var value, keyCode, placeholder, width;
+		var value, keyCode, printable, placeholder, width;
 		var shift, character, selection;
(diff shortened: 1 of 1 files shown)


joallard:
We're discussing a more thorough fix in #809, we need to revise the design of the autogrow method. Thanks for the insight.

johnsenpeder:
:+1:

PR #984, opened 2016-01-12
Add anchors to individual demos

sirlancelot:
I want to be able to link my co-workers to specific demos by URL.

This patch allows adding it to the anchor like so:

http://selectize.github.io/selectize.js/#github-demo


joallard:
That sounds sensible. Would you amend your commit and  put `demo-` at the beginning of the id so there's a sense of hierarchy in them?

sirlancelot:
Updated

joallard:
Fantastic, thank you.

PR #1015, opened 2016-03-04
Add dropdownDirection option

ftalves:
Based on ideas discussed in #29.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -140,6 +140,12 @@ $(function() {
 		<td valign="top"><code>false</code></td>
 	</tr>
 	<tr>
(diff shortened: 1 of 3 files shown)


joallard:
There's already #437 open, maybe you can merge any improvements with them?

ftalves:
I read the solution in that pull request, didn't actually test his code but mine appears to be more clean and simple in terms of code, while achieving the same thing . What should I do in this case?

joallard:
I'm afraid we'll need to process their pull request before doing yours, and see if we can improve upon that. The project needs significant time just to be managed, I'm sorry I can't help more directly with that for now.

Issue #733, opened 2015-03-10
Required fields can not be focusable


oswaldoacauan:
Error: `An invalid form control with name=''" is not focusable.`
`/examples/required.html`
Browsers: `Chrome v40` and `Chrome Canary v43`
SO: `Windows 7`

Ref.: #63 and #136

Seems that the issue is back :sweat_smile:

scaytrase:
Confirm, required inputs become hidden and unfocusable (which leads to submit fail). Also I cannot make the selectized input required.

conarwelsh:
same here 
![image](https://cloud.githubusercontent.com/assets/2643188/6716169/20101cfa-cd69-11e4-8a49-178ead73d7a1.png)

therocode:
Also experiencing this

rudymacias:
Same here.

seanodotcom:
Seeing this in Chrome 42 as well.

chibicode:
Same here on Chrome 43.

Alex-Sokolov:
Some sort of workaround...

Let style invalid selects with red border

``` css
.selectize-control.invalid > .selectize-input {
    border-color: red;
}
```

And in init of each select add code for checking validation

``` js
$('select').each(function() {
    var $select = $(this).selectize(options);
    selectizeValidationFix($select);
});

function selectizeValidationFix($select) {

    var self = $select[0].selectize;

    self.$input.on('invalid', function (event) {
        event.preventDefault();
        self.focus(true);
        self.$wrapper.addClass('invalid');
    });

    self.$input.on('change', function (event) {
        if (event.target.validity && event.target.validity.valid) {
            self.$wrapper.removeClass('invalid');
        }
    });

}
```

Result: no error message in console about "not focusable" and styling invalid selects with red border. But if on page few selects - all opened together

dexcell:
Yeah, hopefully this bug is going to fixed soon.

I did a quick dirty fix since i also need it. At least it worked for me.
On selectize.js (un-minified) line 2655.

Change the code from

```
self.$control_input.prop('required', invalid);
```

 to 

```
self.$control_input.prop('required', self.$input.val() == '');
self.$input.prop('required', false);
```

Use at your own risk. Only tested aganist dropdown style of selectizejs.
Hopefully it help someone.

oswaldoacauan:
@brianreavis any thoughts on this?

sintro:
Same problem, "is not focusable" problem is still here

[...]

PR #1029, opened 2016-03-24
Fix unfocusable required select fields #733

ricobonfim:
A lot of people has been changing their own version of selectize.js to fix this problem. With this fix, I suppose most of them can start pointing to the official repository once again.

This fix was suggested by @dexcell

### DIFF
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -1600,7 +1600,8 @@ $.extend(Selectize.prototype, {
 		var invalid, self = this;
 		if (self.isRequired) {
 			if (self.items.length) self.isInvalid = false;
-			self.$control_input.prop('required', invalid);
+			self.$control_input.prop('required', self.$input.val() == '');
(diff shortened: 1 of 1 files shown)


joallard:
Can you explain what was the problem? Also, are you able to write some tests? Thanks

fuhrysteve:
I believe this is to fix the error in some browsers: `An invalid form control with name='fieldname' is not focusable`

Basically to reproduce you need to have something like this:

``` html
<select required>
    <option value="">select an option</option>
    <option value="1">option1</option>
</select>
```

Now selectize it, and try to submit the form. If I didn't miss anything, you should get the above error.

Here's what happens when you combine `required` with `value=""` (to have a placeholder): the browser tries to focus on the hidden select element to say "this field is required" (since `""` is selected), but since it can't focus on a hidden element, it raises a javascript error.

PR #1035, opened 2016-04-05
Fix bug in group order

### LINKED ISSUE #703: lockOptgroupOrder breaks onChange
Using selectize 0.12.0, with `lockOptgroupOrder: true`, firing an onChange event returns the following error:
**TypeError: t.optgroups[b] is undefined** on line 2129 `var b_order = self.optgroups[b].$order || 0;`

wizzvet-cborsenberger:
Fix bug with groups_order


joallard:
Reffing #703

PR #1054, opened 2016-04-28
avoid calling createItem twice. fixes #1001

### LINKED ISSUE #1001: Hitting tab when creating a new entry causes the create function to trigger twice
If on a dropdown types in a new entry ("Test") and then hits Tab rather than selecting "Add Test" two new entries are created.
`

```
$('#transaction_payee_id').selectize({
    allowEmptyOption: true,
    sortField: 'text',
    selectOnTab: true,
    createOnBlur: false,
    closeAfterSelect: true, 
    placeholder: "Select payee",
    create: function(input, callback) {
        $.ajax({
            type: "POST",
            url: "/payees",
            data: { payee: { name: input } },
            success: function(data){
                console.log(data);
                return callback({ text: input, value: data.id});
            },
            error: function(){
                alert("Ajax error!");
                    return false;
            }
        });
    }
});
```

`

If I set "selectOnTab: false" the user is still able to create a new entry ("Test") even if "Test" already is an option.

kdiogenes:


winzig:
I know it's two years later ;-) but this also fixes the same exact problem for me. Any chance this full request is going to get picked up?

PR #1056, opened 2016-05-04
Sugestion for keepPlaceholder variable

nordqvist:
Conditions in the updatePlaceholder function to keep the placeholder visible until the maximum or all available items has been set. Controlled by the variable keepPlaceholder. Default to false.


joallard:
Is there an issue related?

Issue #882, opened 2015-08-20
Pasting text with the mouse doesn't trigger search


demon101:
For selectize.js with ajax search inserting text by mouse not cause search

It's can be simle reproduced on http://brianreavis.github.io/selectize.js page. On Remote Source — Github example:
1. focus on field
2. delete selected
3. insert text any text by mouse (not by ctrl+v)
4. no result

How to fix it?

For catching event by jquery bind method. Selectize on method can't catch it (bug?).

$('.selectize').bind('input', function(){
// force selectize to make ajax call and show result  
});

// following code catch nothing
$('.selectize')[0].selectize.on('input', function(){
// force selectize to make ajax call  
});

But can't find solution for force selectize ajax call and show result

demon101:
Bug fix is

```
onPaste: function(e) {
            var self = this;
            if (self.isFull() || self.isInputHidden || self.isLocked) {
                e.preventDefault();
            } else {
                // If a regex or string is included, this will split the pasted
                // input and create Items for each separate value
                 setTimeout(function() {
                     if (self.settings.splitOn) {
                        var splitInput = $.trim(self.$control_input.val() || '').split(self.settings.splitOn);
                        for (var i = 0, n = splitInput.length; i < n; i++) {
                            self.createItem(splitInput[i]);
                        }
                      }
                      self.onKeyUp(e);
                }, 0);
            }
        },
```

Jaza:
The fix from @demon101 works for me, I've put it in a pull request.

Syone:
It seems this issue isn't solved yet

risadams:
closing stale issues older than one year.
If this issue was closed in error please message the maintainers.
All issues must include a proper title, description, and examples.

PR #1066, opened 2016-05-20
Make mouse pasting trigger search

Jaza:


Jaza:
PR for fix provided by @demon101 in issue #882

joallard:
Ref #882

src/selectize.js
```diff
@@ -406,14 +406,15 @@ $.extend(Selectize.prototype, {
 		} else {
 			// If a regex or string is included, this will split the pasted
 			// input and create Items for each separate value
-			if (self.settings.splitOn) {
-				setTimeout(function() {
+			setTimeout(function() {
+				if (self.settings.splitOn) {
```
joallard, line 410: Know the rationale for this?

kerhong: I believe it's just to make `self.onKeyUp(e)` (like 461) always be triggered inside the delayed function. The issue is not with order of `setTimeout` and `if`, but in the fact that `self.onKeyUp(e)` also needs to be triggered after a delay.

Without `self.onKeyUp(e)` there is no code inside this function that actually handles the data.

winzig:
@Jaza Why was this pull request closed? The issue still happens for me. Was a fix attempted in another pull request?

winzig:
Still using this pasting fix to this day, and it's working. Would be awesome to get this merged.

PR #1073, opened 2016-06-04
Add `dropdownDirection` option (supports `up` and `down`).

jaridmargolin:
Many of the core concepts here come from https://github.com/selectize/selectize.js/pull/437/files. Thank you @streamside for the original work. We have stripped out `auto` functionality, so that these pieces may be implemented in more manageable chunks.

A few thoughts:
- The bootstrap styles I had to write were just overrides of the new styles I declared in selectize.less. Ideally there would be a core selectize stylesheet with required/shared styles, and then additionally theme styles. This would remove the necessity for these overrides. (Sits outside of the scope of this PR).
- QA was rather manual due to varying themes. Karma appears to be well suited for testing the functionality of selectize, but lacks the necessary capability to catch visual "breaks". You could possibly take screenshots and diff the images (may be the only way to have a high degree of confidence in visual consistency accross browsers)?
- Adding new styles (like drop for example), still requires the human eye to confirm it is behaving correctly. In the case of this feature I created examples/direction.html to confirm visuals behaved correctly across themes.
- A new test file was added, `positioning.js`. It should prove useful for future work on `dropdownDirection: auto`... But I do not want to note, it is not perfect. The tests are based on my understanding of how things are/should be positioned. The tests are only as good as my understanding. If you have thoughts on a more robust solution, I'd be happy to implement.

---

All in all, I have a high degree of confidence no breaking changes were implemented, and that the new functionality is working as expected across each theme.  While I believe, due to the complexity in the changes made, that everything is behaving correctly across browsers, I cannot guarantee it because I did not manually check each browser.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -140,6 +140,12 @@ $(function() {
 		<td valign="top"><code>false</code></td>
 	</tr>
 	<tr>
+		<td valign="top"><code>dropdownDirection</code></td>
+		<td valign="top">Direction in which dropdown will display. Currently accepts "up" and "down" as possible values.</td>
+		<td valign="top"><code>string</code></td>
+		<td valign="top"><code>down</code></td>
+	</tr>
+	<tr>
 		<td valign="top"><code>dropdownParent</code></td>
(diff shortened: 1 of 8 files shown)


joallard:
That looks great; great work @jaridmargolin! Let me look this over, getting this merged is in my priorities for the project.

PR #1080, opened 2016-06-14
clearOptions: Don't remove already selected items

### LINKED ISSUE #1079: How to dynamically update options on type
We are using selectize to allow a user to add hashtags to a post. When the widget is initialized, I make a request to the server that populates the dropdown with trending hashtags. Then, as the user types, I would like to update the options to show tags that match their search. e.g. (this is used in Ember)

```
let s = input.selectize({
   searchField: ['name'],
   preload: true,
   options: [],
   create: function(input, callback) {
   },
   render: {
      option: function (item, escape) {
         return '<div>' + escape(item.text) + '</div>';
      }
   },
   load: (query, callback) => {
     if (!query.length) {
        return this.initializeFromTrendingTags().then((options) =>  {
           // returns an array of options [{ text: ..., value: ... }, ...]
           callback(options);
        });
     }
     // searchTags also returns an array of matched options in format [{ text: ..., value: ... }, ...]
     return this.searchTags(query).then((options) => {
        callback(options);
     });
  }
})
```

Where searchTags issues the api request and updates the options. 
However, no matter what I do I can't seem to get the options to update. Ideally I'd like the dropdown to change while open to the matched tags. I've tried doing the following:

```
selectize.clearCache('option');
selectize.clearOptions();
selectize.addOption(options);
selectize.refreshOptions(true);
```

but this just clears out the options altogether. I'm stuck. any help? 
thanks!!!

andrey-lizunov:
Closed #1079

### DIFF
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -1336,10 +1336,15 @@ $.extend(Selectize.prototype, {
 		self.loadedSearches = {};
 		self.userOptions = {};
 		self.renderCache = {};
-		self.options = self.sifter.items = {};
+		var options = self.options;
+		$.each(self.options, function(key, value) {
+			if(self.items.indexOf(key) == -1) {
+				delete options[key];
+			}
+		});
+		self.options = self.sifter.items = options;
 		self.lastQuery = null;
 		self.trigger('option_clear');
-		self.clear();
 	},
 
 	/**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+* Fixed bug making `clearOptions` function. Now it doesn't remove already selected options. (thanks @caseymct - #1079)
 * Fixed bug making `allowEmptyOption: true` useless (thanks @mcavalletto - #739)
 * Functions in option `render` can now return a DOM node in addition to
   text. (#617)


joallard:
Can you explain your PR? I wasn't able to understand the issue ref'd.

andrey-lizunov:
@joallard I found that selectize accumulates options from previous searches. I upload new list of options remotely from my server and the number of selected items not limited. In this case after several searches in dropdown I have a lot of options connected with old queries which I don't need.
I tryed to call `clearOptions` before upload new list of options, but `clearOptions` calls `clear` function inside and this removes all items which was chosen before.
The problem that I want to be able to clear just list of options without removing already selected items. I talking about multiple selects.

This PR change this behavior, now the list of selected items stay unchanged, but list of options for select is cleared.

src/selectize.js
```diff
@@ -1336,10 +1336,15 @@ $.extend(Selectize.prototype, {
...
 		self.userOptions = {};
 		self.renderCache = {};
-		self.options = self.sifter.items = {};
+		var options = self.options;
+		$.each(self.options, function(key, value) {
+			if(self.items.indexOf(key) == -1) {
+				delete options[key];
+			}
+		});
+		self.options = self.sifter.items = options;
```
brookjordan, line 1345: Is this not the same as:

``` javascript
$.each(self.options, function(key, value) {
  if(self.items.indexOf(key) == -1) {
    delete self.options[key];
  }
});
self.sifter.items = self.options;
```

mtebele:
Any workaround for this?

src/selectize.js
r2my: @brookjordan I confirm, your code works fine.

colalex:
Is this still open? 
I have the problem only on Android (other system work fine even with chrome) when I want to delete a selected item.

andrey-lizunov:
@colalex This issue is not about deleting a selected item.
It's about clearing (generate new list) options when you already have selected options and your list of options are dynamic (ajax from server).

ravins:
still open ?

kobykorman:
@joallard Any chance you can merge this? A couple of these really small fixes for impactful UX issues are hanging around, and we could really use your help.

joallard:
Had to rebase that one

kstratis:
2021 and this is still an issue... unbelievable...

risadams:
> 
> 
> 2021 and this is still an issue... unbelievable...

Would you believe that a PR closed two years ago might not be quite relevant to today's environment?  Feel free to open a new issue with a working example, or submit your own PR with the changes you wish to see.

PR #1092, opened 2016-06-26
Get rid of Error: Invalid US-ASCII character "\xE2"

ashatrov:
To get rid of error when I insert CSS generated from this LESS into mine SASS file.
Error: Invalid US-ASCII character "\xE2"


joallard:
Files are in UTF-8 format, not ASCII. `0xE2 0x80 0x93` is the correct sequence for char `U+2013` En-dash. Make sure whatever is consuming the file understands properly UTF-8.

Issue #1098, opened 2016-07-01
Highlight only works for the first typed letter


skimi:
The highlight only highlights the first or first and second letterz (if you are fast) of the input text.
Afterward anything you type is ignored.

Also, if you copy paste a string in the input it will be highlighted completely but will ignore anything you type after that string.
This must be related to some kind of timing issue.

![highlight](https://cloud.githubusercontent.com/assets/10855180/16522751/e26e3260-3f9f-11e6-8f47-b2a655a85971.png)

**Steps to reproduce:**
1. On http://selectize.github.io/selectize.js/ Email Contacts example
2. Type "brian"
3. Only "b" or "br" is highlighted

**Expected result:**
Brian should be highlighted in both the email and the name

**Actual result:**
Only "b" or "br" is highlighted

**Might be related:**
The highlight `<span class="highlight">` gets wrapped inside other `<span class="highlight">` at each new highlight instead of resetting the string.

skimi:
Ok, I've debugged a bit more.

The bug is due to the way the highlight function splits the string to insert the `<span class="highlight">`. After the split, every letter is on its own line and is considered a full word by the function.

In [newer version of the highlight function](http://johannburkard.de/blog/programming/javascript/highlight-javascript-text-higlighting-jquery-plugin.html) they have added a removeHighlight() function that solves the problem by reverting all highlighted strings to their original state.

I would simply update the highlight script to the last version but the last version has changed it's matching system and it does not pass the js strict mode so it can't be commited as is into selectize repo.

Anyway I've made a minimal commit where I've just updated selectize version to include their removeHighlight() and edited it to pass strict mode.

0xbaigorria:
Hey guys!

I bet this is already fixed since Jonathan closed it, but I wanted to add to what Skimi pointed out, that a second issue is that since highlights don't get cleared, you end up with multiple `<span>`s wrapping the same or different highlighted letters. Check screenshot.

![selectizehighlightbug](https://cloud.githubusercontent.com/assets/4060852/17036556/26fe78e0-4f63-11e6-89b3-7557cb436bf3.png)

_I'm using Safari Version 9.1.2 (11601.7.7) on OS X 10.11.6 (15G31)._

skimi:
Yep it'll be fixed in 0.12.3. Previous highlights will always get cleared before new ones are added.

Larivact:
[...]

PR #1099, opened 2016-07-02
Fix highlight not working after first highlight insert #1098

skimi:
The highlight did not work correctly after the first highlight happened. Most likely it would only highlight the first letter.

[Newer version of the highlight function](http://johannburkard.de/resources/Johann/jquery.highlight-5.js) have added a removeHighlight() function that allows to reset the strings between each highlights and then getting a good result.

Problem, that version does not use Regexp to match the highlight anymore and their file does not pass the strict mode as is so it can't be commited to selectize.

It's a bit weird to have that contrib commited into the repo. I don't know if it's wise to edit it in our code or not or to update it or to remove it, or... So I've just made the minimal changes to fix the issue without changing much. I've copied their removeHighlight() fn and edited it to pass strict mode.

From issue #1098

### DIFF
--- a/src/contrib/highlight.js
+++ b/src/contrib/highlight.js
@@ -37,3 +37,16 @@ var highlight = function($element, pattern) {
 		highlight(this);
 	});
 };
+
+/**
+ * removeHighlight fn copied from highlight v5 and
+ * edited to remove with() and pass js strict mode
+ */
+jQuery.fn.removeHighlight = function() {
+	return this.find("span.highlight").each(function() {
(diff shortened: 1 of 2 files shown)


joallard:
Thanks, I'll look it over!

groovenectar:
I tried this patch. In my case it seems to work well even on a ~4900 item list. Thanks!

vedmaka:
:+1: please merge, encountered this issue recently too

joallard:
Merged thanks

PR #1105, opened 2016-07-08
Improve performance of large option lists

StefanH:
I've noticed large option lists in selects cause the browser to timeout when rendering selectize. 
Detaching the input, then detaching the children from it, 
then re-attaching the input dramatically improves performance.

### DIFF
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -225,8 +225,13 @@ $.extend(Selectize.prototype, {
 
 		// store original children and tab index so that they can be
 		// restored when the destroy() method is called.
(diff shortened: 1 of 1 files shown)


joallard:
Thanks for the patch.

So I've tried this out with a new example in the performance demos and a profiler open. I see no difference with a select with 12k options. Firefox is `5.5s -> 6s` and Chrome 51 `2.5s -> 2.5s`

Would you try branch `pr-1105` at `examples/performance.html`, 2nd example, and figure out if there's a way to test this? Thanks!

PR #1106, opened 2016-07-09
Remove keypress checks to fix autogrow jitter for some characters.

Billiam:
Addresses issue where common, non-(a-z/0-9) characters jogged left until a key was released.

https://github.com/selectize/selectize.js/issues/809

This PR uses the input event to avoid printable character checks entirely. Instead, just check the field value after something's been entered/removed/pasted/etc.

This breaks IE8/9 compatibility, so a [polyfill](http://benalpert.com/2013/06/18/a-near-perfect-oninput-shim-for-ie-8-and-9.html) may be required. This may not be acceptable, depending on long-term roadmap.

selectize

Updated demo page here: https://billiam.github.io/selectize.js

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -271,41 +271,13 @@ var autoGrow = function($input) {
 	var currentWidth = null;
 
 	var update = function(e, options) {
-		var value, keyCode, printable, placeholder, width;
-		var shift, character, selection;
+		var value, placeholder, width;
+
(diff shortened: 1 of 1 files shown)


joallard:
Quite elegant. Since this breaks IE8/9, let's classify this as a breaking change and schedule it for 1.0. If we can integrate a polyfill, I'd consider moving it forward.

PR #1109, opened 2016-07-15
Add new plugin, select_on_focus

christor:
This plugin clears the selection, but leaves the label for the selection in the text edit field. For convenience, the text is fully selected so a user can begin typing to replace the current value easily.


christor:
Just noticed that this is similar in spirit to #776 (the basic idea is to improve UX by allowing a user to start typing after focusing on a selectize field). This just exposes it in a slightly different way, selecting the text to give the user a cue that they can type over what's there.

I think I actually like #776 a bit better but I notice that it hasn't been merged. Is there something preventing merge? I'll close this since I like the other solution better.

PR #1112, opened 2016-07-20
No need to remove the selected item matching if there is no input

ghoshnirmalya:
Return false from `removeItem` if there is no input.

### DIFF
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -1487,6 +1487,10 @@ $.extend(Selectize.prototype, {
 		var self = this;
 		var $item, i, idx;
 
+		if (!value) {
+			return false;
+		}
+
 		$item = (value instanceof $) ? value : self.getItem(value);
 		value = hash_key($item.attr('data-value'));
 		i = self.items.indexOf(value);


joallard:
Thanks. Can you explain a short example where that case would happen?

ghoshnirmalya:
I'm using this Ember plugin (https://github.com/miguelcobain/ember-cli-selectize). 

So, whenever I'm using the select box in a page and going to the next route, it throws me this error `Uncaught TypeError: Cannot read property 'attr' of null`. I found out that the issue was coming from this line: `value = hash_key($item.attr('data-value'));` when the `value` argument in this line `removeItem: function(value, silent) {` becomes `null`.

joallard:
Okay, interesting, and who calls `removeItem` with a `null`?

ghoshnirmalya:
This line calls it: `self.trigger('option_remove', value);`

joallard:
Line 1343 in `removeOption`, `self.trigger('option_remove', value);`, does not call `removeItem`.

ghoshnirmalya:
this line (https://github.com/selectize/selectize.js/blob/master/src/selectize.js#L1335) calls it.

jakeonfire:
@ghoshnirmalya i got that error too with `ember-cli-selectize` when trying to use `content.name` for the `optionValuePath`. i changed it to `content.id` and no longer get that error. might have something to do with your computed `optionValuePath`s changing? hope that helps. (actually created a mixin that adds a `selectizeId` to any model so that i can also use newly created items which caches a random-ish ID for selectize's use)

PR #1117, opened 2016-07-27
Add openOnDelete setting

nimajneb:
Hi,

I quickly needed an option to prevent the results dropdown from being automatically opened when an item was removed, so here it is. I hope it's useful to others!

Cheers,
Benjamin


joallard:
Thanks for the PR. I'm going to leave it as is for the moment, but as many reports about the dropdown automatically opening have outlined, this area is problematic and needs to be refactored a little more deeply.

PR #755, opened 2015-04-02
Make label[for] work with selectized input

barretts:
this updates any existing labels pointing to the original input or select to point to the selectize input/select #658

maintained id uniqueness by appending '_selectized' string

PR #1119, opened 2016-07-28
Properly escape CSS notation

fkaempfer:
Small problem caused by #755  as it does not escape CSS notation within the id attribute properly. Compare: https://learn.jquery.com/using-jquery-core/faq/how-do-i-select-an-element-by-an-id-that-has-characters-used-in-css-notation/

### DIFF
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -129,7 +129,8 @@ $.extend(Selectize.prototype, {
 
 		if(inputId = $input.attr('id')) {
 			$control_input.attr('id', inputId + '-selectized');
-			$('label[for='+inputId+']').attr('for', inputId + '-selectized');
+			var inputIdEscaped=inputId.replace( /(:|\.|\[|\]|,)/g, "\\$1" );
+			$('label[for='+inputIdEscaped+']').attr('for', inputId + '-selectized');
 		}
 
 		if(self.settings.copyClassesToDropdown) {


joallard:
Can you point to an example id where that would be the case?

fkaempfer:
Sure. One of the most popular Java web frameworks (JSF) creates IDs separated by ":" depending on their position in the DOM, for example "myForm:mySelectField". The ":" causes an exception in the label selector code.

joallard:
``` gherkin
Given I have a label and input field with id 'form:name'
When I selectize this field
Then the resulting field should have id 'form:name-selectized'
And the label should have attr for 'form:name-selectized'
```

As I understand, that's not currently the case?

joallard:
Got it. That said, there's an even easier fix: http://stackoverflow.com/q/9763323/720164

``` diff
- $('label[for='+inputId+']').attr('for', inputId + '-selectized');
+ $("label[for='"+inputId+"']").attr('for', inputId + '-selectized');
```

I can let you commit that if you want the credit ;o)

fkaempfer:
Great, yeah that is much simpler. I checked that it works in my case.

No need to give me credit, but if it makes it easier for you to integrate the fix I can update the PR.

Thanks!

joallard:
Fixed, thanks!

PR #1125, opened 2016-08-04
Removed the non-ASCII '–' character within the license part which causi…

hKyo89:
**Problem:**
Found an error when compiling scss on Shopify with the following error message:

`SCSS file is not UTF-8 encoded`

The non-ASCII character resulting Shopify to fail compiling the whole scss files and giving you a _failed-compiled_ output file as shown below on Chrome:

image

image

**Root cause:**
Results of the non-ASCII character found inside the the `less` directory _(as the error occurred only for scss on Shopify)_: 

image

The character:

image

Results of files inside `src` directory containing the character:

image

**Solution:**
Removed the non-ASCII '`–`' em-dash(??) character within the license part of every occurring file at the line of:

`* Copyright (c) 2013–2015 Brian Reavis & contributors`

### DIFF
--- a/src/.wrapper.js
+++ b/src/.wrapper.js
@@ -1,6 +1,6 @@
 /**
  * selectize.js (v@@version)
- * Copyright (c) 2013–2015 Brian Reavis & contributors
+ * Copyright (c) 2013-2015 Brian Reavis & contributors
  *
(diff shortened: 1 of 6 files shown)


joallard:
Dup #1092. The character is valid UTF-8, and compilation using latest Less works.

PR #1126, opened 2016-08-04
Adding config value to search nested fields

ashireman:
Adding this config option will let sifter look through subfields for the search query.

Code below includes some 'searchField' values as an example.

In addition to passing the nesting value to the correct sifter config, this also requires the addition of the 'nesting: true' key and value to your selectize config:

```
{
  searchField: ['field_a', 'field_b', 'some.dot_notation.field_name'], 
  nesting: true
}
```

Whether you want to add this to the base level of the config is up to you. You could also add it to a second level such as:

```
{
  searchField: ['field_a', 'field_b', 'some.dot_notation.field_name'], 
  searchFieldOptions:  [{nesting: true}],
}
```

which would require extra config processing, but might make the placement of those options more clear.

If desired I can likely come up with a short example to add to the documentation as well. It would be worth noting that the subfield searched may also be displayed if highlighting is desired.

### DIFF
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -978,7 +978,8 @@ $.extend(Selectize.prototype, {
 		return {
 			fields      : settings.searchField,
 			conjunction : settings.searchConjunction,
-			sort        : sort
+			sort        : sort,
+			nesting     : settings.nesting
 		};
 	},
 


joallard:
Sounds good, I'll schedule it

anx-ckreuzberger:
Any update when this is going to be in 0.13.0? I'm missing exactly this feature (using nested objects for search), and I've helped myself by forcefully overwriting the function like this:
```javascript
/**
 * Overwrite Selectize getSearchOptions
 * ToDo: Once this pull request gets merged, we do not need this anymore...
 * https://github.com/selectize/selectize.js/pull/1126
 */
$.extend(Selectize.prototype, {
        getSearchOptions: function () {
            var settings = this.settings;
            var sort = settings.sortField;
            if (typeof sort === 'string') {
                sort = [{field: sort}];
            }

            return {
                fields: settings.searchField,
                conjunction: settings.searchConjunction,
                sort: sort,
                nesting: settings.nesting
            };
        }
    }
);
```

PR #1127, opened 2016-08-08
Improve handling under iOS

zeitiger:
Under iOS 
- Open selectize widget with closeAfterSelection: true
- Onscreen keyboard appears 
- Select something (independent of keyboard usage)
- Selectize close correctly, but onscreen keyboard keeps open

This pull request fix this issue, with removing focus on hidden text input field, within close event handling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+*  Fix open keyboard bug under iOS after closing selection (#1127)
+
+   *@zeitiger*
+
 *  Fix highlighting more than one character (#1099, #1098)
 
    *@skimi*
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -1741,6 +1741,7 @@ $.extend(Selectize.prototype, {
 
 		if (self.settings.mode === 'single' && self.items.length) {
 			self.hideInput();
+			self.$control_input.blur(); // close keyboard on iOS
 		}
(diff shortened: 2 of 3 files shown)


joallard:
Thanks. The UX makes a lot of sense to me in the scenario you describe. I'm just wary about adding yet another option. Is there a way we can make the default better?

joallard:
Yep. Actually, having tried this on iOS, having the keyboard still out after selection makes no sense. Let's make this default behavior.

However, the control input must only lose focus in a case like this, where we have a select. When we have an input, the dropdown should close, but the control must not lose focus. I think we'd need to enumerate the use cases to provide solid defaults.

zeitiger:
After I get the broken test, I thought that focus thing was as designed, why else would be assertion for this detail ;-) 

Yes to lose focus within tag mode make no sense, I only use this widget for a improved vanilla select. 

So next step, I kick out the option and check that we are on single mode and made a selection and has `closeAfterSelection: true`. Is that okay for you? 

btw. thanks for your work here, I really appreciate your positiv response 😸

joallard:
Squash that with a concise/descriptive commit message, and we'll be good for a merge

joallard:
Also, don't forget to add a Changelog line!

zeitiger:
done

PR #1128, opened 2016-08-09
Duplicated options in different optgroups doesn't render correctly

zeitiger:
Duplicated options (options with the same value) will be render only once within last occurrence

This pull request includes a test case to reproduce the problem and the bugfix for that. I'm open for discussion to make this bugfix better :-)

Have a nice day 😸

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+*  Fix render bug of options with duplicated values in different option groups (#1128)
+
+   *@zeitiger*
+   
 *  Output friendly error message when Microplguin is missing (#1137).
    Special thanks to @styxxx for proposing the improvement.
 
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -1087,7 +1087,8 @@ $.extend(Selectize.prototype, {
 					groups[optgroup] = document.createDocumentFragment();
 					groups_order.push(optgroup);
 				}
-				groups[optgroup].appendChild(option_html);
+				// a child could only have one parent, so if you have more parents clone the child
+				groups[optgroup].appendChild((!j) ? option_html : option_html.cloneNode(true));
 			}
 		}
 
--- a/test/setup.js
+++ b/test/setup.js
@@ -86,6 +86,24 @@
 					'Group 2': {label: 'Group 2', val: 'Group 2', $order: 4}
 				});
 			});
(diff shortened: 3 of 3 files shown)


joallard:
Thanks for the PR. If I'm not mistaken that should be part of the wider improvement that duplicate values should be allowed regardless of optgroups (see #129). Sounds like those two should probably go hand in hand.

zeitiger:
I believe that is a simple bug, because the complete preprocessing of optgroup and option within `init_select` with `addGroup` and `addOption` had already in mind how I expect that a `select` should work

``` js
// if the option already exists, it's probably been
// duplicated in another optgroup. in this case, push
// the current group to the "optgroup" property on the
// existing option so that it's rendered in both places.
```

and also another test case give me the hint

``` js
it('should allow for values optgroups with duplicated options', function() {
```

but that fall short, because the DOM works different as expect for the original author within rendering step.

The issue want a change within the data processing step before the rendering. If you want that that you have to change the internal data structure, that will be a much bigger refactoring. 

IMHO my pull request is as render bug fix independent of data processing change request

zeitiger:
This already existing example (`optgroups.html`) is IMHO broken at the moment

``` html
            <div class="demo">
                <h2>Optgroups (repeated options)</h2>
                <div class="control-group">
                    <label for="select-repeated-options">Options:</label>
                    <select id="select-repeated-options" class="demo-default" multiple>
                        <option value="">Select...</option>
                        <optgroup label="Group 1">
                            <option value="a">Item A</option>
                            <option value="b">Item B</option>
                        </optgroup>
                        <optgroup label="Group 2">
                            <option value="a">Item A</option>
                            <option value="b">Item B</option>
                        </optgroup>
                    </select>
                </div>
                <script>
                $('#select-repeated-options').selectize({
                    sortField: 'text'
                });
                </script>
            </div>
```

current version render this version 
bildschirmfoto 2016-08-11 um 13 07 30

and my change will make it render this way
bildschirmfoto 2016-08-11 um 13 08 16

zeitiger:
I made an entry in the changelog and rebase this pull request to make it easier to merge :-D

joallard:
I will check it out, thanks @zeitiger

yaquawa:
Same problem here, still not merged yet!

joallard:
If I could get a rebase on master, I'd be inclined to merge.

PR #1131, opened 2016-08-11
feature: allow to disable single options or complete optgroups

zeitiger:
Selectize.js at the moment doesn't respect vanilla HTML `option` and `optgroup` flag `disabled`

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option

This pull request include demo and implementation. I know that tests are missing, I will fix this tomorrow. 

I only want feedback about this and timezone ping pong, before a put more effort in this. 

Have a nice day
David

PR #1133, opened 2016-08-12
Define npm standard commands within package.json

zeitiger:
In reference to https://github.com/selectize/selectize.js/pull/1131
I split my changes within build chain from the other pull request and move it in this one
- start local environment with `npm start`
- build default version with `npm build`
- run test suite with `npm test`, already existed command but improve because you don't need full relative path
- grunt npm script alias, this has the benefit you don't need a global installed grunt, `npm run grunt -- --plugins=remove_button,restore_on_backspace` will do the same like `grunt --plugin=remove_button,restore_on_backspace`

This are only aliases and are only in interest for new developers follow the npm script philosophy of unified build commands

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ $ BROWSERS=Firefox,Chrome,Safari npm test
 To run Selectize locally:
 
 ```sh
-$ grunt serve
+$ npm start
 ```
 
 You can then run the examples in `http://localhost:8000/examples/`.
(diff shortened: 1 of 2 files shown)


zeitiger:
As good instruction for the npm script stuff I recommend to watch https://www.youtube.com/watch?v=zWEU8kNKi3Q

PR #1135, opened 2016-08-12
Feature request: Override control input type

zeitiger:
Currently the control input is a simple text input. 

Sometimes its quite helpful override the input type for touch devices e.g. for input of numeric data is a `<input type="tel">` better option for a small smartphone.

image1

This pull request include 
- 2 test cases (select and input)
- documentation for a new setting
- 2 new demo cases with input-tag-with-dialpad and select with search country telephone codes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+*  Feature to allow override input control type for e.g. keyboard suggestion on touch devices (#1135)
+
+   *@zeitiger*
+   
 *  Output friendly error message when Microplguin is missing (#1137).
    Special thanks to @styxxx for proposing the improvement.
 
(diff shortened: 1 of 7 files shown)


joallard:
Interesting, I like it. On an `<input>` tag, shouldn't we mirror the original type by default? Meaning if I selectize an `input[type=tel]` tag, it should be applied to the control.

zeitiger:
Sure is this possible, the problem was that browser like Chrome response to something like that

``` html
<input type="number" value="5,23,42,1337"/>
```

with an error because `5,23,42,1337` is not a number on the HTML level. 
So in the most cases it will be the better option to say the source input element is a plain text input and override type of control input within user settings. 

I add now the logic take input type from source input as suggestion for type of control input and add some test cases for that.

zeitiger:
IMHO I would prefer `inputmode` attribute, but that isn't supported by any browser at the moment

``` html
<input inputmode="tel">
```

http://caniuse.com/#feat=input-inputmode

PR #1150, opened 2016-09-10
Improve IE11 fix not to throw errors when element cannot be focused

mpokrywka:
Sometimes element passed as second argument to onBlur handler
doesn't have focus method, i.e. inline svg element in IE


joallard:
I'll take it.

PR #1151, opened 2016-09-15
Fix remove_button plugin when settings.maxItems=1

Pictor13:
As explained in the comments here https://github.com/selectize/selectize.js/pull/848#issuecomment-247317001, I fixed the library in the case the setting _maxItems_ is equal to "1".

All the test pass correctly.

Please evaluate if merging this pull-request, after verifying that nothing else breaks (everything seems fine to me).

Moreover, **it still has to be compiled!**. I didn't do it because the compilation also changes other parts of the code that I am not sure are supposed to be changed (and I wonder also why those changes happen; different dependencies versions maybe?).

An example is that the [Line 1133](https://github.com/Pictor13/selectize.js/blob/master/dist/js/selectize.js#L1133) is changed to 

> `dest && dest.focus && dest.focus();`

and am not sure _what_ makes the change (grunt task?) and _why_.
Maybe you can take care of compiling and making a new release?

Please take a look at it.
Thank you.


Pictor13:
@joallard the last pulled fix (afd0c9c ) to the _remove_button_ plugin had some issues.
I fixed it (explanation in the commit message), fixed an additional bug with event triggering, and supported an additional option.

Please, verify the pull-request.

joallard:
Indeed, you must not change the dist folder, and only put your changes in the `src` folder. Changes are compiled into dist every time there's a release.

To get this merged:
- Remove your changes in `dist`
- Squash your commits into one

Thanks!

Pictor13:
@joallard I was able to revert the changes to the _dist_ folder to the **v0.12.4**.

But I already pushed all the single commits to my remote origin; squashing them would mess up the history (am I missing something here? Is there another way?).
Btw, each commit addresses a different issues; I think it makes sense to keep them divided.

Please let me know how to proceed.

joallard:
Ouch. I thought I'd check it out myself and do some Git-fu on it, but it's really tangled up bad. Which commits are yours?

Some commits that were already in the master branch were re-added to your branch, another merge with master was made, then there's the weird commit on the tip (the "restored" one). 

If you need help visualizing, I use `gitk --all`. Ideally what we're looking for is to have your changes in a series of commits branching off from the main branch.

Pictor13:
Ok, I tried again ^^

I close this one; **correct pull-request is here *#1311***.

Didn't have the time to fix my fork master yet, so for now there is just a new *fix* branch.

Current *upstream/master* has been integrated into the *fix* branch on my fork, and the commits with the fixes have been added on top of that (hopefully in a more clean and less messy way than with this pull-request).

I think now should be clear and ready to be merged without issues.

Sorry for the initial mess with my contribution, and the terrible delay in fixing it.
Hope it will help! :)

PR #1152, opened 2016-09-20
Remove Single don't add close arrow

tchiotludo:
The arrow to remove an item on single select is always missing on current version.
The RemoveSingle plugins return 2 html elements (the initial one and the clsoe arrow) and the render only the first one 

``` js
    render: function(templateName, data) {
            // ...
            html = $(self.settings.render[templateName].apply(this, [data, escape_html]));
            // ...

            return html[0];
        },
```

I also fix the layout that is broken (the arrow was over the text) for all themes


IcyEagle:
Thanks for that @tchiotludo !

joallard:
I'll take you at your word

PR #1163, opened 2016-09-29
RefreshOptions: Add truthiness check when sorting optgroups with lockOptgroupOrder

bwilson-ux:
When using 'lockOptgroupOrder' - if b is not defined it throws an error breaking the component. This change fixes this issue by checking whether it has something to compare first before trying to get access to the object property.


joallard:
Oops, you changed `dist`!

bwilson-ux:
@joallard thanks for the heads up! I only updated dist to rectify my current issue not thinking about the actual build... Updated patch also :+1:

joallard:
Sorry to make you go back and forth on this, but there's no way to do this simply in the Github interface: can you rebase and remove your first commit, leaving just the changes to `src/`? 

Otherwise I'm ready to merge as-is, it's trivial enough to be comprehendable.

PR #1166, opened 2016-10-03
Add typescript support

jantimon:
This pull request allows typescript codebases to use selectize:

``` js
import selectize from 'selectize';
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,625 @@
+// Type definitions for Selectize 0.12.13
+// Project: https://github.com/brianreavis/selectize.js
+// Definitions by: Adi Dahiya <https://github.com/adidahiya>, Natalie Bausch <https://github.com/naBausch>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+/// <reference types="jquery"/>
+
+declare namespace Selectize {
+    // see https://github.com/brianreavis/selectize.js/blob/master/docs/usage.md
+    // option identifiers are parameterized by T; data is parameterized by U
+    interface IOptions<T, U> {
+
+        // General
+        // ------------------------------------------------------------------------------------------------------------
+
+        /**
+         * The string to separate items by. This option is only used when Selectize is instantiated from a
+         * <input type="text"> element.
+         *
+         * Default: ','
+         */
+        delimiter?: string;
+
+        /**
+         * Enable or disable international character support.
+         *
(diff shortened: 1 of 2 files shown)


joallard:
I welcome the proposition, can you provide a little more background about this? Personally, I don't know Typescript at all (except by name).
1. Can you do a little background about how it fits into a Typescript project and how it works?
2. Any thoughts about whether this could be an external plugin?
3. As this enters the codebase, it will have to be maintained. What does this entail from maintenance?
4. How will this have to change as the Selectize code base changes?

Feel free to provide more info as needed.

jantimon:
Typescript is the microsoft version of babel - it is a ES6/ES7 to ES5 transpiler.

But there is more than just ES6 and ES7 it also adds typing (that's why it is called typescript)

As you see in the merge request I added a `index.d.ts` file. It's actually only a description of your plugin and explains which methods exist, which parameters you can pass to the function (e.g. the options object) and what they return.

This allows to get error messages if you add call selectize with an unkown option: 
components_atoms_type-select_js_type-select_ts__fix_selectize_

It would be cool if it would be part of your codebase as it would be installable using npm and the version of the type definitions would always match with your current api.
So if you would decide to remove an option typescript would tell us that we are using an "unknown" option after we updated your plugin and we can adjust our code.

Maintainence:

The index.d.ts has only to be updated if a new method or option is added but it is really simple now that the base file is already there and you have a good example how to extend it.
I can also assist you if you need help with it.

Do you need more information on one of the topics?

jantimon:
One Example why it might be good for your library is that Google is using Typescript for Angular2.
So in order to use selectize in an Angular2 app those definitions are required.

joallard:
Thanks for the writeup. All right, I'll schedule it for next minor.

jantimon:
Cool 👍 

Can you already give an estimation when this 0.13.0 will be released?

joallard:
Yet I'm not able to, I don't have the means to spend time maintaining the project unfortunately. It's done on free time.

jantimon:
Why do you create such big releases instead of using semVer - it would be so much simpler and less work to release after every pr and have small change sets instead of gathering large releases.. also people would be motivated to help you even more if they get the feeling that they are involved.. and you could also add a text to the readme "partial maintained looking for full time maintainer" so people know.

(i am maintaining an module in my spare time which is used intensively)

risadams (approved):

PR #1176, opened 2016-10-19
Hide autocomplete on Android

ulriklystbaek:
Issue with browsers internal autocomplete gets shown in chrome on some Android devices. Seems to be that autocomplete is not respected when missing name. I have taken the attributes google uses on their input for google main page, and it seems to work.


joallard:
Okay for the attributes, not sure for the name, there could be collisions.

ulriklystbaek:
You are right, I went a bit fast. I have also gotten to know selectize alot better now, I will try to make a better solution later. But now I am finishing full Android support (backspace and keypress problems), and adding the ability to use other than 'text' also changing it after creation.

PR #1194, opened 2016-11-04
Preservation of original data attibutes and banned values in search

fran-diaz:
Added variables and code to make use of banned values and preservation of option's original data attributes. In my case, i fill bannedValues array externally to filter values refered by selection on other select input. Example:

```javascript
$('.selectize').selectize({
        preserveOptionsAttr: true,
        preload: true,
        onChange: function(value){
            var linked_input = this.$input.attr('data-linked-field');
            var linked_value = this.options[value]['data-linked-value'];
            var banned_values = [];
            
            if(linked_input !== undefined && linked_value !== undefined){
                $('#'+linked_input)[0].selectize['bannedValues'] = [];
                ops = $('#'+linked_input)[0].selectize.options;
                for(var opt in ops){
                    if(ops[opt]['data-linked-value'] !== linked_value){
                        banned_values.push(ops[opt]['value']);
                    }
                };
                
                $('#'+linked_input)[0].selectize['bannedValues'] = banned_values;
                $('#'+linked_input)[0].selectize.refreshOptions(false);
            }
        }
    });
```

### DIFF
--- a/dist/js/selectize.js
+++ b/dist/js/selectize.js
@@ -1506,7 +1506,7 @@
 			}
 	
 			// perform search
-			if (query !== self.lastQuery) {
+			if (query !== self.lastQuery || (self.bannedValues !== undefined && self.bannedValues.length >= 1)) {
 				self.lastQuery = query;
 				result = self.sifter.search(query, $.extend(options, {score: calculateScore}));
 				self.currentResults = result;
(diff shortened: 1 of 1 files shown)


joallard:
I'm not clear on what behavior you're fixing/enhancing here, can you explain the scenario and maybe add a test if it's important?

PR #1197, opened 2016-11-11
Add "better_auto_grow" plugin which improves autoGrow performances

marcotisi:


joallard:
Thanks, I'll try to review that. However, if the performance is just better, we should merge in to the main source instead of just having that as a plugin. Would you be able to make a diff over what we have right now?

marcotisi:
Yeah, tomorrow i'll push a PR with changes directly to the core.

PR #1199, opened 2016-11-12
Improve measureString performance

marcotisi:
Create only once a <span> element and use it to measure string.
Remove element if Selectize.count equal 0.

### DIFF
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -2049,6 +2049,11 @@ $.extend(Selectize.prototype, {
 		self.$control_input.removeData('grow');
 		self.$input.removeData('selectize');
 
+		if (--Selectize.count == 0 && Selectize.$testInput) {
+			Selectize.$testInput.remove();
+			Selectize.$testInput = undefined;
+		}
+
 		$(window).off(eventNS);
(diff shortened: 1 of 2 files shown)


marcotisi:
@joallard this is the PR which improve measureString performances integrated in the main source.

marcotisi:
Any chance to merge this PR in a short time? @joallard

src/selectize.js
```diff
@@ -2049,6 +2049,11 @@ $.extend(Selectize.prototype, {
 		self.$control_input.removeData('grow');
 		self.$input.removeData('selectize');
 
+		if (--Selectize.count == 0 && Selectize.$testInput) {
```
joallard, line 2052: What this condition does is not immediatly clear

joallard: Can you break this out in a better descriptively named function?

joallard:
Do you have benchmark stats?

src/selectize.js
marcotisi: Yeah, I can break into a function and give some benchmarks in a couple of days.

PR #1201, opened 2016-11-15
Fix special char issue in firefox

topaxi:
I've noticed that in Firefox with newer versions of jQuery (version 2 or 3), that special characters like `@`, `-` or even umlauts like `ä` won't be inserted into the select box.

I've tracked down the bug to the `measureString` function, I don't know what is _really_ causing the bug, but if I remove the `$test` element asynchronously, it seems to be working in Firefox again.

Might be related to #809.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -253,7 +253,7 @@ var measureString = function(str, $parent) {
 	]);
 
 	var width = $test.width();
-	$test.remove();
+	setTimeout(function() { $test.remove(); }, 0);
 
 	return width;
 };


src/utils.js
```diff
@@ -253,7 +253,7 @@ var measureString = function(str, $parent) {
 	]);
 
 	var width = $test.width();
-	$test.remove();
+	setTimeout(function() { $test.remove(); $test = null }, 0);
```
ThiefMaster: The `$test = null` doesn't seem to be very useful here

ThiefMaster:
This issue seems to be caused by the `body` element using `flex` display. Here's a codepen with a minimal example reproducing it: http://codepen.io/ThiefMaster/pen/zNVJRB

It smells like a Firefox bug but this workaround would indeed help.

Getting a bugfix release containing this fix would be very useful to our project since we have users complaining about this problem.

ThiefMaster:
https://bugzilla.mozilla.org/show_bug.cgi?id=1341030

topaxi:
Since Firefox has patched the issue on their side I'm going to close this :)

Issue #1204, opened 2016-11-16
Selectize deletes data attributes from select options


kodeo:
When selectize is applied to a select, its options' data attributes are removed.

```
<select id="s">
  <option value="foo" data-bar="baz" selected>Foo</option>
</select>
```

```
var $s = $("#s");
alert($s.find("option:selected").data("bar"));
$(s).selectize();
alert($s.find("option:selected").data("bar"));
```

https://jsfiddle.net/u6qobn24/

kodeo:
Have fixed this in PR #1206, although I now see this fails a test. Think the test may need rewriting. Maybe someone else can assist.

hajikelist:
this is fairly important, any movement on this?

ghost:
+1 to this. I think all data-attributes on select and options should be preserved.

planotransmedia:
+1 to this. as well

Oxicode:
+1

dokle:
+1

Oxicode:
@joallard @brianreavis please merge

CristianSitov:
+1

joallard:
@Oxicode This is an issue, it's not mergeable

vivekapex:
+1

dantheother:
+1

tonksullan:
+1

paulzi:
+1

ptlergo:
+1

coughsyrupgit:
+1

ansien:
+1

risadams:
closing stale issues older than one year.
If this issue was closed in error please message the maintainers.
All issues must include a proper title, description, and examples.

Oxicode:
Everything you should not do.

> closing stale issues older than one year.
> If this issue was closed in error please message the maintainers.
> All issues must include a proper title, description, and examples.

PR #1206, opened 2016-11-17
Preserve original options

kodeo:
Store original options and use in updateOriginalInput to preserve data attributes etc. re: #1204.


CedNet:
Hi, please prioritize this as it'd really improve the easy of use by a lot! Thanks

Oxicode:
+1

kodeo:
I rescind this pull request as it's been too long now and I think the change needs to be made by someone who understands the library better than I do anyway. 

I still think preserving the data attributes is a very important feature to add though.

Issue #1164, opened 2016-09-30
Hitting Tab won't jump to next field (regression in 0.12.3)


sturchyn:
Hi,

the behaviour on KEY_TAB is broken in the version 0.12.3.

Here examples to reproduce it:

selectize 0.12.2 (works)
https://jsfiddle.net/sturchyn/b961qfjk/

selectize 0.12.3 (doesn't work)
https://jsfiddle.net/sturchyn/vyggj6oz/

(Given a Selectize with option `selectOnTab:true`)

Steps to reproduce:
1. Set focus on the select box.
2. Select the option 'B'. 
3. Press the key 'TAB'.

Expected result:
- The option 'B' is selected.
- The select box is closed.
- The cursor jumps to the next input field.

Actual result (with 0.12.3):
- The option 'B' is selected.
- The select is open.
- The focus is on the select box.

Many thanks for Selectize :-)

Best regards
Stanislav

joallard:
Indeed, great report, thanks!

hupf:
This regression has been introduced by the following change in #1127:
https://github.com/selectize/selectize.js/commit/d7015c9f9f7b218c2c82988ba19756295b1a3698#diff-6cee306935e9c39b0f577d4ccb04fadeR1744

If you comment this line, the tab behavior works as expected.

Any idea how we can fix the iOS issue without breaking the default tab behavior?

joallard:
@hupf If you get around to a PR, do mention me, I've checked out of notifications but this seems like a priority.

hupf:
Here's my pull request: https://github.com/selectize/selectize.js/pull/1215

joallard:
Merged

copperx:
The issue is *still* present in 0.12.4 when using the following options

```
            create: false, 
            preload: true,
            selectOnTab: true,
            sortField: 'name',
```

and having `onChange`, `render`, and `load` callbacks.

I've gone back to 0.12.2 to avoid the issue.

Elias-Serneels:
Just commenting to confirm this issue in 0.12.4 on firefox and IE using following options:
```
closeAfterSelect: true,
maxItems: 1,
persist: false, // setting this to true has same problem.
create: someFunction
```

reverting to 0.12.2 solved this particular issue but one issue on firefox remained: tabbing from a selectize field to a regular input field kept focusing back to the selectize field.

This last issue was solved by incrementally setting the tabindex of each input field every time a form is shown:
`
[...]

PR #1215, opened 2016-12-05
Fix tab key regression introduced by touch handling #1164

hupf:
Unfortunately [You Can't Detect A Touchscreen](http://www.stucox.com/blog/you-cant-detect-a-touchscreen/) but we can escape the call stack (using ``setTimeout``) before bluring. That allows the default tab key action to take place first.

I've tested it in Firefox and Chrome (Desktop and Android) -- I was able to reproduce #1127 on Android too, but I don't have an iOS device at hand...

### DIFF
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -1752,7 +1752,9 @@ $.extend(Selectize.prototype, {
 
 		if (self.settings.mode === 'single' && self.items.length) {
 			self.hideInput();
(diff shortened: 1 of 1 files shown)


hupf:
Was able to test it on an iOS device as well, it works fine.

hupf:
@joallard This is the solution I came up with...

joallard:
We'll try that, thanks!

PR #1226, opened 2016-12-29
Dynamic option group/optgroup implementation

### LINKED ISSUE #151: Loading optgroup results from the server.
It seems awkward having to manually specify the potential optgroups for selectize to properly group items. I understand that the current implementation offers a little more flexibility in terms of optgroup labels/ids, but when loading search results from a server it would be super handy if selectize would automatically pull out all the unique optgroups from the results, perhaps only when the optgroups array was not set manually during initialization.

I've tried to set the optgroups array within the load event (after pulling out the unique optgroups), but that doesn't seem to be working.

### LINKED ISSUE #1025: Callback does not currently support dynamically loaded option groups?
I'm looking through the source code at the load event and it does not seem that the callback function passed to load supports loading option groups from the server. Is there plans to implement this in the future? 

I'm looking around line 1767 for the load function in selectize.js.

jackbentley:
Fixes #1025 (I think?)
Fixes #151


risadams (approved):

risadams:
Updated and merged in #1912

PR #1227, opened 2016-12-29
Allow the dropdown to reopen on click (closeAfterSelect)

pedantic-git:
When `closeAfterSelect` is true, the selectize doesn't lose focus, which means it doesn't reopen when the user clicks it again unless they take the focus away first. This feels unnatural to me (because it's different from how a `<select>` behaves).

I've made this little change that allows the `onClick()` handler to run if the dropdown has been closed by `closeAfterSelect`.

Let me know if you'd rather I made this optional, but it seems pretty natural to me and doesn't break any of the existing tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 <!-- Feel free to put either your handle and/or full name, according to
      your privacy needs -->
 
+*  Allow the dropdown to reopen on click if it is closed without losing focus
+   by closeAfterSelect: true
+   
+   *@fishpercolator*
+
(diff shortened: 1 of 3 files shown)


pedantic-git:
Hi selectize maintainers - do you have any comments about the PR? I think it's hopefully a simple bugfix and it would be great to get it accepted upstream if you don't have any issues with it.

kobykorman:
@joallard This bug is alive and kicking and seriously major UX-wise, and the fix is super light on LOC changed, any chance you can give this priority and merge?

PR #1251, opened 2017-02-20
Use JQuery.detach on removeItem so custom data attributes are retained

john3parker:
I'm using a custom Item renderer with some custom data attribute fields. When .remove is used, JQuery removes the data-* and event handlers. Since selectize caches the rendered item it doesn't get created each time so the data-* attributes and even handlers are lost. To prevent his situation I changed the .remove() to .detach(). I also added a test to check for data-* attribute retention.


giero:
It's kind of connected with my issue: https://github.com/selectize/selectize.js/pull/1315

Issue #479, opened 2014-06-17
FastClick and Selectize.js on iOS 7, can't select from dropdown


kinergy:
Just posted [this bug](https://github.com/ftlabs/fastclick/issues/265) on the FastClick repo - 

FastClick is a useful utility that eliminates the 300ms delay when taping to click in iOS. Currently though, a user can't select an option from a Selectize dropdown in iOS when FastClick is enabled.

I just wanted to see if anybody in this group has any ideas about what may be causing the issue.

Thanks!

adamgins:
@biesbjerg thanks for this. 

I am still having issues selecting on a touch device in both Safari & Cordova,

I can select an item / option in the list. I am using the `create=true` option and I can't `add` a new item by clicking on it. If I hit enter it works.. .it's just the click.

I tried overriding the `option_create` too with the following function:

```
function(item, escape) {

     return '<div class="create needsclick">Add <strong>' + escape(item.input) + '</strong>&hellip;</div>';
};

```

But same symptom: nothig happens, it just deselects the dropdown like nothing was clicked. Not sure what I am doing wrong.

adamgins:
Ok, I found a workaround `createOnBlur:true,` . I think when I touch the "add something... " it's closing the dropdown. So the blur event  cateches this.. so I am not sure how good a workaround it is.. but feels like a bug.. that said I am using touchpunch ... which may be part of the issue here.

AdamWills:
I'm having the same issue as @adamgins here.

farturi:
I'm having the same issue as @adamgins ...

danielgehr:
We where able to solve the problem @adamgins describes by adding the touchstart event to the selectize. Check out this PR: https://github.com/selectize/selectize.js/pull/1262

adamgins:
Thanks  will take look.

risadams:
closing stale issues older than one year.
If this issue was closed in error please message the maintainers.
All issues must include a proper title, description, and examples.

PR #1262, opened 2017-03-17
Issue 479 - add touchstart in order to catch touches as well as clicks when adding custom items

danielgehr:
We faced the same problem with selectize on iOS as _adamgins_ describes in this issue:
https://github.com/selectize/selectize.js/issues/479

Basically we where not able to add new custom items to the selectize by clicking the _"add-option"_. It worked fine when running on a desktop machine using the mouse or when adding the option on an iOS device by hitting _enter_ on the keyboard. Just clicking the _"add-option"_ on a touch device didn't work.

Seemed like somewhere the touch got lost while the click was respected.
We where able to solve that problem by adding the `touchstart` event to the selectize.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 <!-- Feel free to put either your handle and/or full name, according to
      your privacy needs -->
 
+*  Bugfix: add touchstart in order to catch touches as well as clicks when adding custom items
+
+   *@danielgehr*
+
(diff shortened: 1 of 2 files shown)


danielgehr:
This change breaks scrolling. Problem is that when using touchstart that the add optionSelect handler gets executed right away without waiting if the user eventually moves his finger in order to scroll. Therefore this change should not be merged...

PR #1265, opened 2017-03-20
Fix broken documentation

siegy22:


brianreavis:
Strange. Thanks for the workaround!

PR #1279, opened 2017-04-27
Adding ability to use load to init opt groups

brainded:
I noticed the ability to init the list from a async source was missing the ability to also init groups, so I added what I thought was needed to get this working. It utilizes the optgroupValueField to figure out the Id to use with the addOptionGroup method and iterates through the list of groups to add them.

I looked for any examples of tests for the load function but didn't find any. I am going to take a look at adding a test once I figured out how to use this testing pattern with the function but I wanted to get this pull request open so its at least on the board.

### DIFF
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -740,11 +740,17 @@ $.extend(Selectize.prototype, {
 		var $wrapper = self.$wrapper.addClass(self.settings.loadingClass);
 
 		self.loading++;
-		fn.apply(self, [function(results) {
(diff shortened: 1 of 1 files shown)


4levels:
Hi,

any prediction on if this will be merged and if so, when?  Seems like a great addition for the project!

Kind regards and thanks again for making selectize so great!

Joozty:
It would be great if this could be merged.

pierreminiggio:
I'd need that. Has this not gotten merged yet? Is there another good alternative?

PR #1296, opened 2017-06-07
Performance improvements

roblevintennis:
These are some performance improvements we've made over last couple of years for the Mavenlink application's use of selectize. We had intended to port these over in to a PR some time ago but got swamped. That said, I suppose they've been more battle tested as a result. Here are the relevant changes:

* Use document fragment to improve performance when adding multiple items d5785f0
* Only position dropdown once when adding multiple items 8f46e77
* Only update original input once when adding multiple items 1782fb4e

### DIFF
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -1502,7 +1502,10 @@ $.extend(Selectize.prototype, {
 
 				self.updatePlaceholder();
 				self.trigger('item_add', value, $item);
-				self.updateOriginalInput({silent: silent});
+
+				if (!self.isPending) {
+					self.updateOriginalInput({silent: silent});
+				}
 			}
 		});
 	},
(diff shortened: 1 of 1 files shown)


roblevintennis:
@coveralls so are you saying I need to have test coverage for those lines of code first or is this an auto-generated message? Sorry if asking obvious, just want to be "crystal clear"; especially since looking at https://coveralls.io/builds/11874223/source?filename=src%2Fselectize.js the methods we changed seem to have coverage (unless it caused others not to be traversed)

brianreavis:
Great changes! I don't really maintain selectize anymore, but I'll try to get a new version out in the next few days – I hope.

You can ignore @coveralls. It's just a bot that seems to be acting up...

roblevintennis:
@brianreavis ahhhh ok sweet! No problem on the new release, I'll probably just point our package json to our fork until you do and then make the switch once it's in. Thanks for the fast response!

PR #1303, opened 2017-06-26
clear_button plugin

Koloto:
The plugin adds a small cross on the right which clears all items (while `remove_button` plugin removes only one item).

### DIFF
--- a/src/plugins/clear_button/plugin.js
+++ b/src/plugins/clear_button/plugin.js
@@ -1,0 +1,52 @@
+/**
+ * Plugin: "clear_button" (selectize.js)
+ * Copyright (c) 2017 Alexey Kolotovchenkov & contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this
+ * file except in compliance with the License. You may obtain a copy of the License at:
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+ * ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ *
+ * @author Alexey Kolotovchenkov <alkoloto@hotmail.com>
+ */
+
+Selectize.define('clear_button', function(options) {
+	var self = this;
+
+	options = $.extend({
+		label     : '&times;',
+		title     : 'Clear',
+		className : 'clear'
+	}, options);
+
+	self.setup = (function() {
+		var original = self.setup;
+		return function() {
+			original.apply(self, arguments);
+
+			var html = '<a href="javascript:void(0)" class="' + options.className + '" tabindex="-1" title="' + escape_html(options.title) + '">' + options.label + '</a>';
+
+			self.$control.append(html);
+
+			// add event listener
+			self.$control
+				.on('mousedown', 'a.' + options.className, function(e) {
+					// prevent toggling
+					e.preventDefault();
+					e.stopPropagation();
+				})
+				.on('click', 'a.' + options.className, function(e) {
+					e.preventDefault();
+					e.stopPropagation();
+					if (self.isLocked) return;
+
+					self.clear();
+					self.close();
+				});
+		};
+	})();
+});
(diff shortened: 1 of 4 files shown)


Koloto:
@brianreavis, @joallard Any chances to merge this PR?

src/plugins/clear_button/plugin.js
```diff
@@ -0,0 +1,52 @@
...
+					// prevent toggling
+					e.preventDefault();
+					e.stopPropagation();
+				})
+				.on('click', 'a.' + options.className, function(e) {
+					e.preventDefault();
+					e.stopPropagation();
+					if (self.isLocked) return;
+
+					self.clear();
```
Pictor13, line 47: *clear()* will not trigger the '*item_remove*' event (see pull request: https://github.com/selectize/selectize.js/pull/1311/commits/ea2e931fc1d1cb4eea530f951206f12401e80042#diff-a1e04fdc9fa0888b83d5b5401328081bR62)

Maybe it is on purpose, but just wanted to mention it, in case *deleteSelection()* is more appropriate.

Pictor13:
- I added a comment to the code, about the use of *clear* method (https://github.com/selectize/selectize.js/pull/1303/commits/a5d61d2398f851ab709dd0b480d2f8e57f75c6a5#r129585962). Just in case..
- Also I am not sure about the reason for changing the *selectize.js* source (https://github.com/selectize/selectize.js/pull/1303/commits/a5d61d2398f851ab709dd0b480d2f8e57f75c6a5#diff-6cee306935e9c39b0f577d4ccb04fadeR1814).

Except for that: thank you for the plugin :)

Koloto:
@Pictor13 
* `deleteSelection()` removes the selected items only, not all. You can select an item by clicking on it (in multi mode), then it will be marked with `active` CSS class. I think `clear` method should be changed to trigger `item_remove` event.
* Changes in `selectize.js` are required to prevent deleting clear button itself.

Pictor13:
* Regarding the ```selectize.js``` change, are you sure it is safe?
I didn't check, but I was thinking that putting the button inside ```selectize.$wrapper``` could make more sense, and would prevent from touching the library code (plugins should serve that purpose).
Just a suggestion of course. If it doesn't break any use case then your change is totally fine.

* About the ```item_remove``` event I wondered the same; but I was focus on fixing the *remove_button* plug-in and I didn't investigate further, taking for granted the bug was in the plugin and not in the main library (evil assumptions...).
I wonder if my pull request is wrong then, in relying on ```deleteSelection()``` for both modes (single/multi). It was the only way I could get all the expected behaviours (and the events triggering).
But I have to admit it felt like a hack to make it work that way; while maybe the fix of ```clear()``` is the right thing to do here, as you wrote.

Also, feel free to comment back my pull request, if you think is wrong! 👍

src/plugins/clear_button/plugin.js
Pictor13: Actually `clear()` already triggers the event *clear*. And that might be enough.

Having the *item_remove* event being triggered automatically would be just convenience; probably one should listen `on("clear item_remove")` in case of wanting to execute the same callback for both events.
E.g. using the same callback for settings `onClear` and `onItemRemove` will achieve that.

Additionally, `deleteSelection` provides support for the `onDelete` callback, that is an actual useful feature (validate values before removal and eventually prevent the clear/remove operation).
I believe this should be still supported also when clearing or removing items (so that we can stick to what said in the previous paragraph).

Pictor13:
I would add some insights (useful for my pull request and for this one as well), coming from reading code and tracing the commit history:

There are some evidences, in the repo history, that `deleteSelection()` is adequate for removing items in any situation (in fact it is [what's being called, regardless the `mode` when pressing DELETE/BACKSPACE](https://github.com/selectize/selectize.js/blob/5b06be3d5726c89ed1dbf2ff22eb94b9f2bc60ff/dist/js/standalone/selectize.js).
In general selection won't be available via UI but can still be set programmatically and is not incompatible with, e.g., `single` mode.

Also, `deleteSelection` supports the `onDelete` callback (the only way in the library to validate values and prevent deletion/removal), while `clear`/`removeItem` don't take that in account.
I believe this to be an incongruence, since specifying an `onDelete` callback **is a feature that should always be available**, rather than being called exclusively in `multi` mode or in a context of a selection.

**Note:** this is not directly related to this pull request, but I wanted to leave a trace for when the above mentioned details will be implemented/fixed.

Issue #547, opened 2014-08-25
Make Single Select Look like Tags


steelbrain:
It would be great to have an option that allows us to make the single select box look more like an input box, tags input for example.

Pictor13:
Was it not documented on purpose?
(maybe because of some grey-area use-case, where combination of settings break, when forcing the `mode`?)

Or can I do a pull-request with `mode`-setting added to the documentation?

Issue #124, opened 2013-09-21
Add support for "single" mode to "delete_button" plugin


brianreavis:
Place an "x" to the left of the dropdown arrow for better usability.

Related: #123, #113

jacek213:
+1

stefanmeschke:
+1

tonimoeckel:
+1

namkazt:
still not happen in 2018 👯‍♂️

Pictor13:
Would #1311 make you happy?

I chose to keep the 'x' close to the items, since atm there is a bug that allows to load multiple values even when the `mode = 'single'`.
You can see that bug trying to use `awsome,neat` as input value, in the [_single-select_ example I added to the plugin page](https://github.com/selectize/selectize.js/blob/a7db68ecafa631f6ac211f977e415c5f6f580930/examples/plugins.html#L36).
In that case there would be overlapping multiple crosses, if positioned close to the dropdown-arrow; hence the crosses close to the items.

risadams:
closing stale issues older than one year.
If this issue was closed in error please message the maintainers.
All issues must include a proper title, description, and examples.

Issue #123, opened 2013-09-20
Unselect a single


jackmcdade:
Correct me if i'm wrong, but it would seem there is no way to "unselect" your option in single mode? Is there a way to accomplish this? No rush, but would be really helpful. Thanks!

Pictor13:
The remove-button is there on _single-select_; I fixed it's styling in #1311 .
Eventually using `alllowEmptyOption: true` will also add the dropdown empty-choice for clearing the current value (#163)

Issue #113, opened 2013-09-14
Backspace not fired when cursor after item in Chrome for Android


waddlesplash:
(@joallard edit: 0. Use Google stock keyboard)
1. Go to http://brianreavis.github.io/selectize.js/ in Chrome for Android.
2. Attempt to backspace in the first example -- it won't work. Firefox for Android works as expected with backspace.

This happens for me on:
Android 4.3, Nexus 7 (build `JWR66Y`), Chrome 29 [latest stable].

If you need any more info, please let me know.

bratanon:
Im still having this issue.

rasmuscnielsen:
Just updated to 6.01 on Sony Xperia, and I'm also still experiencing the issue.

joallard:
I see the need, but that's not an easy one. Anyone has a reasonable solution?

rasmuscnielsen:
@joallard I made this workaround on the angular-selectize directive: https://github.com/machineboy2045/angular-selectize/pull/139

It allows for an easy way to add a special 'clear option' in the dropdown itself. When selected, it will clear the input.
For the directive, the name of the option is added as an attribute, so for instance it can be 
`<input type='text' selectize options='[]' clear-option='Any'>`

I guess a possibility like this in the selectize library itself would not only solve the problem, but be a nice feature?

![skaermbillede 2016-06-09 kl 21 53 57](https://cloud.githubusercontent.com/assets/8465957/15944266/1ac48f16-2e8d-11e6-92ab-39d77fd63bd4.png)

joallard:
Seeing [chromium#118639](https://bugs.chromium.org/p/chromium/issues/detail?id=118639) was closed as `wontfix` in Google's _infinite wisdom_...

@rasmuscnielsen I'm sorry, I'll have to disagree with the UX here. What I could see as viable workarounds however, is that one use the RemoveButton plugin, or have a `X` on the right side of the field.

joallard:
I've fooled around on https://w3c.github.io/uievents/tools/key-event-viewer.html a bit to see what we could do for an ugly workaround.

Here's the event log of what happened:

```
# 1. Click field, type X

- keydown 229, value ''
- input, 'X'
- keyup 229, 'X'

# 2. Type 'Y'

- keydown 229, 'X'
- input, 'XY'
- keyup 229, 'XY'

# 3. Backspace Y

- input, 'X'
- keydown 229 'X'
- keyup 229 'X'

# 4. Backspace X

- input, ''
[...]

PR #1311, opened 2017-07-24
Fixes to remove button plugin

Pictor13:
This replaces the pull request I made for fixing the *remove_button plugin*, cause of mess I made with the commits (changed the *dist/* and not just the *src/*; and used ugly revert commit to repair).

I didn't squash the commits as each concerns a different fix/detail.

Please refer to pull request  "#1151 *Fix remove_button plugin when settings.maxItems=1*" for details.
I am going to close that one, referring this pull request.

PS - Sorry for taking so much time to recreate it. Hope it will help somebody.

### DIFF
--- a/examples/plugins.html
+++ b/examples/plugins.html
@@ -31,6 +31,10 @@
 					<label for="input-tags2">Tags:</label>
 					<input type="text" disabled id="input-tags2" class="input-tags demo-default" value="awesome,neat">
 				</div>
+				<div class="control-group">
+					<label for="input-tags">Tags:</label>
+					<input type="text" id="input-tags3" class="input-tags2 demo-default" value="awesome">
+				</div>
 				<script>
 				$('.input-tags').selectize({
 					plugins: ['remove_button'],
@@ -44,6 +48,14 @@
 					onDelete: function(values) {
 						return confirm(values.length > 1 ? 'Are you sure you want to remove these ' + values.length + ' items?' : 'Are you sure you want to remove "' + values[0] + '"?');
 					}
+				});
+
+				// single selection
+				$('.input-tags2').selectize({
+					plugins: ['remove_button'],
+					persist: true,
+					create: true,
+					maxItems: 1
 				});
 				</script>
 			</div>
--- a/src/plugins/remove_button/plugin.js
+++ b/src/plugins/remove_button/plugin.js
@@ -62,7 +62,9 @@ Selectize.define('remove_button', function(options) {
 						e.preventDefault();
 						if (self.isLocked) return;
 
-						self.clear();
+						// use removeItem - clear() will not trigger the 'item_remove' event
+						var $item = $(e.currentTarget).parent();
+						self.removeItem($item);
 					});
 
 				};
--- a/src/plugins/remove_button/plugin.js
+++ b/src/plugins/remove_button/plugin.js
@@ -24,7 +24,7 @@ Selectize.define('remove_button', function(options) {
 
 		var singleClose = function(thisRef, options) {
 
-			options.className = 'remove-single';
+			options.className = options.className || 'remove-single';
 
 			var self = thisRef;
 			var html = '<a href="javascript:void(0)" class="' + options.className + '" tabindex="-1" title="' + escape_html(options.title) + '">' + options.label + '</a>';
@@ -37,8 +37,8 @@ Selectize.define('remove_button', function(options) {
 			 * @return {string}
 			 */
 			var append = function(html_container, html_element) {
-				return $('<span>').append(html_container)
-					.append(html_element);
+				var pos = html_container.search(/(<\/[^>]+>\s*)$/);
+				return html_container.substring(0, pos) + html_element + html_container.substring(pos);
 			};
 
 			thisRef.setup = (function() {
--- a/src/plugins/remove_button/plugin.js
+++ b/src/plugins/remove_button/plugin.js
@@ -16,111 +16,111 @@
 
 Selectize.define('remove_button', function(options) {
 	options = $.extend({
-			label     : '&times;',
-			title     : 'Remove',
-			className : 'remove',
-			append    : true
-		}, options);
+		label     : '&times;',
+		title     : 'Remove',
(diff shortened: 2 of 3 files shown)


anx-ckreuzberger:
I just tried your "remove_button" and the "x" icon just magically appeared, without having to change any of my configuration. Thank you so much!

temuri416:
Not sure this is working as intended for the case when settings.maxItems=1.

If you edit 

https://github.com/selectize/selectize.js/blob/master/examples/plugins.html#L38

and add **maxItems: 1** to the config, your selectize control will look like this:

image

Pictor13:
## Update
Had no time to check all the little details and the consequences; sorry if this PR stalled.
Atm I am trying to:
* check the bugs or little disfunctions coming with this PR
* check what was the original logic intended by the author (regarding the combination of the implicitly dependant settings: `maxItems`, `mode`, `allowEmptyOption`)
* check what the ChoppyThing's PR broke of that logic and original feature (IMO it broke in different points)
* confirm all the use-cases are ok before confirming this PR
* (eventually) add unit-test or add some documentation about the insights that I get to

### Bugs
* in a project of mine the PR fix *seems* to provoke another bug:
selecting and then removing an item, got from a query A (with `load`), won't be shown in the dropdown, when trying to reload (typing) the same query A; running a query that was not previously cached, and then trying again the previous steps, seems to restore the missing result in the dropdown.
Not sure it is related, but *maybe* the plugin rendering steps on the rendering called by `load`.
* with `maxItems: 1` or `mode: single`, providing two or more values in initialisation (as value of the original HTML element) seems to load them (see plugin-example page), rather than skipping the too many items.

### ToDo
* 🔳 verify if it happens always and find a fix.
* ☑️ check bug reported by @temuri416 and find a fix
* * 🔳 verify why remove-button seems to not appear, in single-mode, with *0.12.3* while it does with the *0.12.6*
* 🔳 verify everything works in multi/single mode and in multi mode with `maxItems=1`
* 🔳 verify the plugin setting `append: false`
* 🔳 verify the points reported in [ChoppyThing's pull-request comments](/selectize/selectize.js/pull/848#issuecomment-320642284)
* 🔳 (if possible/enough time) add unit-tests

### To understand
(in order to stay consistent and don't have breaking/unintended consequences)

* is `single` mode an implicit `minItems: 1`? (if that setting would exist)
~I think so, according to what a single HTML `<select>` is: at least one option **must** be selected.~ _No; selectize always support the removal of items_
* ~`removeItem` removes the item without checking if it is `single` or `multi` mode or if `allowEmptyOption` is enabled; makes me think that it doesn't matter if a single-select becomes empty (contrary to the previous point).~  _<-- indeed_
~**Todo:** verify behaviour when triggering `removeItem` on a single-select with `allowEmptyOption: false`; then try it with `allowEmptyOption: true`.
**Todo:** in the first case, verify whether the congruency-check for the number of items, in single-mode, is done somewhere else (during rendering/refreshing?) or if it is done at all.~ _(misunderstood)_
* ~is `mode` setting *safe enough*? In #547 is mentioned to not be documented; are there reasons for that? Is it broken/unsafe in some circumstances?~ _`mode` looks safe to me_
* ~shouldn't a control with `maxItems: 1` avoid at all to load more than one item? Atm, as you can see in the plugin-examples page, it won't be able to select more than one value, but it is initialised with the two values provided (*neat*, *awesome*); shouldn't the rendering/refresh methods ensure to not render more items than specified in `maxItems`? Why 2 items are rendered (it should just have 2 options and 1 item: the first or the last)~
_This is a bug, but not of concern with this pull-request_

Pictor13:
@temuri416 that's how it is supposed to look like.

### Why
It changes visually because, when you just specify `maxItems: 1`, [*Selectize* falls back to **single** mode](/selectize/selectize.js/blob/v0.12.6/dist/js/standalone/selectize.js#L1202) so like a normal single `<select>`.
(also the CSS will be for single, so no styling for items as cards/boxes, since there is just a single choice possible).
If you remove the `render` setting, that wraps the item-value with `"`-quotes, it will look like as expected (single select).

screen shot 2018-10-07 at 20 21 10

**Note:** the position of the "x" is maybe a little misaligned, but it is not a bug. Maybe the *less* file should be adjusted.

### Solution
If you wanna get the cards/boxes styling you need to manually force the **multi** mode via the setting `mode: 'multi'` when initialising.

(I didn't notice because in my project I tend to always specify the desired mode rather than making it implicit by combination of settings)

### By the way....
* ~I am not sure forcing the `mode` is completely safe though: [@brianreavis didn't add the setting to the documentation](/selectize/selectize.js/issues/547) so I am not sure if this is supported and working in the 100% of cases.~
_probably a useless concern; using `mode` looks fine to me_
* ~Also to notice that Brian's code, initially, [was preventing to have the *remove-button* in **single** mode](/selectize/selectize.js/blob/500a7d6138b2865a9ab596d11000250b6ed48a2a/src/plugins/remove_button/plugin.js#L18).
And it makes sense to me, since in that mode the behaviour should be the same as a normal html-single-`<select>` (at least one option **must** be selected). But this [has been changed arbitrarily by @ChoppyThing](/selectize/selectize.js/commit/856307cfd603b84fa457b261343de48ab6f27228#diff-a1e04fdc9fa0888b83d5b5401328081bL18) in her pull-request, that I am trying to fix~
 _Wrong, it is still possible to remove the choice using **`[backspace]`**_
* _doesn't matter; I misunderstood what that setting does._
~the *remove_button* plugin was developed **before** the implementation of the `allowEmptyOption` setting and (I suppose) it was not taken in account at that time.
IMO the **single** mode should support the *remove_button* plugin **just** when `allowEmptyOption` is enabled. Does it make sense to others than me?~

Pictor13:
I fixed the styling. Now looks decent also with single-selects.

Is there something more to fix here, or can we merge?

Pictor13:
When merged this it could close also #124, #123 and #113.

Pictor13:
Referencing the pull-request #1303 because of some insights in its comments, that are useful in evaluating also he changes made in here.

PR #1315, opened 2017-08-10
Add event 'item_before_remove' just before item is removed

giero:
It will be useful to add event just before item is removed, because removed element is detached from the document tree and - for example - it's impossible to get this items' parent.


giero:
It's been a while ;) I would like to refresh this change :)

giero:
@joallard Hi - any chance to merge this change to main repo?

giero:
Is there any chance this could be reopened? I think it's still a valuable change.

databyte:
@risadams - thoughts on reopening and looking at this since change with tests exists? I don't see any negative consequences of adding events to the process.

risadams:
Agreed.  I'm open to reevaluating anything that is closed if there is interest

risadams (approved):

PR #1320, opened 2017-09-04
Proper way to set/unset required attributes: fix for MS Edge

### LINKED ISSUE #1319: Cannot submit form with selectized [required] field in Microsoft Edge
I did:

	  like below

In Microsoft Edge (I could only test it on v14), submitting a form doesn’t work when there’s a selectized `select` inside that has a `required` attribute. Attempting to submit the form simply does nothing in Edge, while it submits the form as expected in other browsers (IE, Firefox, Chrome). I’ve created a minimal test case here: https://codepen.io/anon/pen/MvxGXd. Just open this example in Edge, click on “Empty” an select an item, then click on “Submit”. It won’t work.

Steps to reproduce:

1. Write the following markup:
```HTML
<form action="/api">
  <select name="test1" id="test1" required>
    <option value="">Empty</option>
    <option value="one">One</option>
    <option value="two">Two</option>
    <option value="three">Three</option>
  </select>
  <button type="submit">Submit</button>
</form>
```
2. Add jQuery and Selectize to the page
3. Selectize the `<select>` with
``` JavaScript
$('#test1').selectize()
```
4. Open the page in Edge and select an item in the field
5. Click on “Submit”

Expected result: It should submit the form because it’s valid.

Actual result: Nothing happens.

Removing the `required` attribute on the `<select>` fields makes submitting possible again in Edge. This seems like a weird MS Edge bug. Maybe a workaround for this can be implemented into Selectize. A basic example with a hidden and required select field works in MS Edge (see https://codepen.io/anon/pen/oeVyVM), so it must be something in Selectize that triggers the Edge bug.

TomOne:
Using .prop('required', true) or .prop('required', false) is non-standard and doesn’t work in all browsers, such as Microsoft Edge. This replaces the occurrences with the proper `attr`/`removeAttr` methods.

Now https://github.com/selectize/selectize.js/blob/master/examples/required.html also works with MS Edge.

Fixes #1319.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 <!-- Feel free to put either your handle and/or full name, according to
      your privacy needs -->
+* Fix issues with required attribute in MS Edge (#1320)
+
+    *@TomOne*
 
 *  New feature: allow to disable single options or complete optgroups
 
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -1660,8 +1660,12 @@ $.extend(Selectize.prototype, {
 		var invalid = !this.items.length;
 
 		this.isInvalid = invalid;
-		this.$control_input.prop('required', invalid);
-		this.$input.prop('required', !invalid);
+		invalid
+			? this.$control_input.attr('required', true)
+			: this.$control_input.removeAttr('required');
+		!invalid
+			? this.$input.attr('required', true)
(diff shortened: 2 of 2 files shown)


TomOne:
I seems the Travis CI build failed because the installed Node version is too old.

rhukster:
👍

PR #1333, opened 2017-09-29
No active class on hover

xanderpeuscher:


xanderpeuscher:
Excuse me, this should be a PR on the forked branch.

PR #1334, opened 2017-10-01
hierarchy plugin

the-liquid-metal:
This plugin is generic code to express cascading behavior between controls/fields. the popular sample of cascading behavior is state-city relationship. With this plugin we just need only one control/field, and without additional custom code.


the-liquid-metal:
Is there anyone can help me with this ci error? I am not familiar with node or ci. Am i missing something?

PR #1353, opened 2017-12-18
Fix FF and IE tabbing/blur issues.

### LINKED ISSUE #1164: Hitting Tab won't jump to next field (regression in 0.12.3)

topaxi:
Fixes #1164

### DIFF
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -1772,9 +1772,13 @@ $.extend(Selectize.prototype, {
 
 		if (self.settings.mode === 'single' && self.items.length) {
 			self.hideInput();
-			setTimeout(function() {
+
+			// Do not trigger blur while inside a blur event,
+			// this fixes some weird tabbing behavior in FF and IE.
+			// See #1164
+			if (self.ignoreFocus) {
 				self.$control_input.blur(); // close keyboard on iOS
-			});
+			}
 		}
 
 		self.isOpen = false;


src/selectize.js
```diff
@@ -1772,9 +1772,13 @@ $.extend(Selectize.prototype, {
 
 		if (self.settings.mode === 'single' && self.items.length) {
 			self.hideInput();
-			setTimeout(function() {
+
+			// Do not trigger blur while inside a blur event,
+			// this fixes some weird tabbing behavior in FF and IE.
+			// See #1164
+			if (self.ignoreFocus) {
```
topaxi, line 1779: I've used `ignoreFocus` as it's set in the `onBlur` method, maybe I should use a different, new, boolean flag?

ChurnM4c:
Will these changes be available soon? We could use them too in our current project. 

Cheers

src/selectize.js
joallard: I'm not actively developing the project, I'll have to take your word for it

joallard:
If you could both validate whether the fix is suitable and ready, I'd merge.

denisfl:
It's seems that this fix is OK. Let's merge :) 🚀

nickolay:
This broke the `should close and blur dropdown after selection made if closeAfterSelect: true and in single mode` test added in #1127 (the PR that introduced the regression being fixed here). Anyone knows what's up with it or should it just be skipped for now?

joallard:
@topaxi I've changed the way this operates in 8c76832 while keeping another test case green. As there is no test for this issue, it will be easy to break. Are you able to confirm at least that the issue has stayed fixed?

PR #1363, opened 2018-01-10
Autofill disable possibility

alshenetsky:
The only option to disable autofill in modern browsers is to provide a possibility to set autocomplete="new-password"


alanbarboza:
It really solves the problem! 
Thanks @alshenetsky

Magellanth:
Would be really usefull to include this option in master branch - Chrome autofill ignores autocomplete="off"

bryceadams:
For others with this issue, you could try solving it through a Selectize plugin, like so:

    Selectize.define('no_autocomplete', function (options){
        var self = this;

        self.setup = (function(){
            var original = self.setup;

            return function(){
                original.apply(self, arguments);
                self.$control_input.attr({
                    'autocomplete': 'nope',
                });
            };
        })();
    });

PR #1367, opened 2018-01-29
Updating the default option data.

dairondv1986:
When i was about to set default data for every select in my page, "Selectize" created the items empty but with the value instead; the problem was that the html of the default select's option was not correctly set inside the created item by Selectize's mechanism so thats why I initialize the every item with the default html inside the current processed option's html.


lfdominguez:
Why this is failed when i used and workss great, thanks @dairondv1986

Issue #1403, opened 2018-05-17
Should spellcheck="false" (et. al.) attributes be set by default?


winzig:
I did:

* [x] Search for if my issue has already been submitted
* [x] Make sure I'm reporting something precise that needs to be fixed
* [x] Give my issue a descriptive and concise title
* [ ] Create a *minimal* working example on JsFiddle or Codepen
	(or gave a link to a demo on the Selectize docs)
* [ ] Indicate *precise* steps to reproduce in *numbers* and the result,
	  like below

In the old days, autocomplete="off" was enough to keep browsers from messing with your text while typing in an input field. However, to cover various browser features across platforms, we tend to use this set of attributes:

`autocomplete="off" autocorrect="off" autocapitalize="off" spellcheck="false"`

Otherwise when trying to use selectize we end up with things like this happening:

<img width="167" alt="screen shot 2018-05-17 at 2 54 16 pm" src="https://user-images.githubusercontent.com/127062/40205763-342a2d12-59e2-11e8-8ea7-b1afb1893c71.png">

In a browser such as Safari, start typing a word such as "awful", stopping at "awf". You'll usually see spelling corrections suggested, as shown above, which interferes with the selectize functionality. 

I would propose that these additional attributes be set on the `<input>` fields created dynamically by Selectize, or even better, "inherited" from the source input field if they have been defined in the existing HTML element.

maximelebreton:
Same problem with Chrome Autofill.

Here is a discussion about it: https://stackoverflow.com/questions/15738259/disabling-chrome-autofill

In last versions of Chrome, `autocomplete="off"` is ignored, and needs to be `autocomplete="<anything but on or off>"`

gustavotrott:
> Same problem with Chrome Autofill.
> 
> Here is a discussion about it: https://stackoverflow.com/questions/15738259/disabling-chrome-autofill
> 
> In last versions of Chrome, `autocomplete="off"` is ignored, and needs to be `autocomplete="<anything but on or off>"`

Right! The default autocomplete in Chrome should be changed to some random string.

winzig:
Not sure if this has changed since April, but I'm using Chrome 74 on macOS, and `autocomplete="off"` is still working for me.

jiteshgolecha:
bump

PR #1404, opened 2018-05-18
Added spellcheck to the list of inherited input attributes

winzig:
We now inherit `spellcheck` attributes, just as it inherits the `autocorrect` and `autocapitalize` attributes from input elements. Especially helpful in Safari.

Fixes issue raised in https://github.com/selectize/selectize.js/issues/1403


winzig:
Note that one test was failing, but not due to the changes I made. The test that's failing is detailed here: https://github.com/selectize/selectize.js/pull/1353#issuecomment-386922258

winzig:
Resolved merge conflict. @joallard Any chance we could get this PR merged? It's a fairly simple change that fixes a UI problem on Safari (and probably other) browsers, by inheriting the `spellcheck` attribute of the input element, if included. Example bug in UI without this fix:  

40205763-342a2d12-59e2-11e8-8ea7-b1afb1893c71

joallard:
Can you confirm the tests pass on your end?

Issue #1277, opened 2017-04-12
jQuery 3 Support


jblacker:
This is more of a discussion/information request.  Has anyone tested selectize.js against jQuery 3?  I'm getting ready to upgrade a project from jQuery 2.2.2 to jQuery 3 and this is the only library that I'm not sure about its compatibility.  I plan on doing some testing on my own at some point, but figured I'd create a post here to see what anyone else's findings were and see if we can make selectize.js's jQuery 3 support "offical."

shaneog:
Did you manage to test this @jblacker? I'm in the same position now that you were in.

avranu:
I've tried running selectize on both jquery 1 and 3, and have seen no differences between the two. YMMV.

RaquelPlaza:
I have upgraded to jQuery 3.2.1 recently and when I run my karma tests I get a selectize error: TypeError: undefined is not an object (evaluating 'a.extend') at bower_components/selectize/dist/js/standalone/selectize.min.js:2

Selectize version is v0.12.1

Has anyone else experienced this problem?

vcajes:
with jQuery 3.2.1 and selectiza 0.12.1 I got this error while following the basic example:

`Uncaught TypeError: Sifter is not a constructor`

dansleboby:
Same error on 2.2.3

DenisLabrecque:
I'm upgrading to jQuery 3.5.1 and get the following deprecations: 

- [ ] `JQMIGRATE: jQuery.isArray is deprecated; use Array.isArray`. Looks like calls to `Array.isArray` have been interspersed with `$.isArray` in selectize standalone 0.12.6
- [ ] `JQMIGRATE: Use of number-typed values is deprecated in jQuery.fn.css`
- [ ] `JQMIGRATE: jQuery.trim is deprecated; use String.prototype.trim`

erropix:
@brianreavis @joallard

Can you confirm if this issue is taken into consideration since WordPress 5.6 has been released with jQuery 3 and I see various migration warnings?

With WordPress 5.7, they will completely remove jQuery Migrate and this will break the Selectize library in WP sites.

risadams:
@ERROPiX there should be no issues,  we are using jquery 3.5.1 for local testing and all tests, demos, and examples are working as expected.

erropix:
[...]

PR #1414, opened 2018-06-04
move jquery to peerDependencies

graingert:
https://nodejs.org/en/blog/npm/peer-dependencies/

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "grunt-contrib-uglify": "^2.0.0",
     "grunt-contrib-watch": "1.x",
     "grunt-replace": "^1.0.1",
+    "jquery": "^3.3.1",
     "karma": "^1.3.0",
     "karma-chai": "^0.1.0",
     "karma-chrome-launcher": "^2.0.0",
@@ -50,7 +51,7 @@
     "mocha": "^3.1.0"
   },
   "peerDependencies": {
-    "jquery": "^1.7.0"
+    "jquery": "^1.7.0, ^2, ^3"
   },
   "scripts": {
     "test": "karma start",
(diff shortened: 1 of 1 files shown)


graingert:
Without this, it can cause selectize to try to use a different jquery than the main jquery

graingert:
@joallard currently selictize is locked to jQuery `>=1.7.0 <2.0.0` for npm users. This pr opens that range up to include v2 and v3, and also ensures that only one version of jQuery is installed and used.

graingert:
Fixes: https://github.com/selectize/selectize.js/issues/1277 I think

matason:
Thanks @graingert, this fixes it for me.

Kovah:
This would also solve [this jQuery security vulnerability](https://nodesecurity.io/advisories/328), which is patched in versions >= 3.0.0.

graingert:
Nice, any chance of a release?

joallard:
Done. 0.12.6

Pictor13:
Great!

About jQuery, do we need to keep it in *bower.json* with that version range, for any retro-compatibility reason, or could we even remove it from there and rely on [bower-away](https://github.com/sheerun/bower-away) to move all the dependencies to package.json (Yarn/Npm) ?
Or do it manually, without bower-away.

Correct me if I am wrong, but I see no reason for having two conflicting definitions in bower and npm; is there?

Pictor13:
Looks like there is an issue with forcing the jQuery version to ```^3.3.1``` (#1428).

I guess ```peerDependencies``` wins over ```dependencies``` and doesn't retain the ```^1.7``` compatibility.

joallard:
Phasing out bower is something I agree we should go in the direction of, I'd be receptive to such PRs

PR #1419, opened 2018-06-15
Enhancement - Hide Input

whattafack:
This is related with => https://github.com/selectize/selectize.js/issues/110

- Added option to hide input (optional)
- Prevents to display keyboard on mobile (Iphone tested)
- Will prevent to have filter functionality (cause input is hidden)


simondoescode:
+1

PR #1426, opened 2018-07-12
Fix wrong defaults in usage docs

### LINKED ISSUE #1167: Wrong default settings
from documentation 

```
maxItems    
The max number of items the user can select. 1 makes the control mono-selection, null allows an unlimited number of items.  

DEFAULT = int    1  (but in repo is null)
hideSelected    If true, the items that are currently selected will not be shown in the dropdown list of available options. 

DEFAULT = boolean false (but in repo is null)
```

Please correct this

Pictor13:
closes #1167

### DIFF
--- a/.bowerrc
+++ b/.bowerrc
@@ -1,0 +1,3 @@
+{
+    "registry": "https://registry.bower.io"
+}
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -121,13 +121,13 @@ $(function() {
 		<td valign="top"><code>maxItems</code></td>
 		<td valign="top">The max number of items the user can select. 1 makes the control mono-selection, null allows an unlimited number of items.</td>
 		<td valign="top"><code>int</code></td>
-		<td valign="top"><code>1</code></td>
+		<td valign="top"><code>null</code></td>
 	</tr>
 	<tr>
 		<td valign="top"><code>hideSelected</code></td>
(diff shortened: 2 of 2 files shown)


Pictor13:
@joallard I put the fix for using the bower registry in this pull-request, I think is fine, right?
Or do you need me to make a different pull-request for the two different fixes? (I guess the difference would be only in my branches name).

[Travis-CI was failing](https://travis-ci.org/selectize/selectize.js/builds/403025826?utm_source=github_status&utm_medium=notification) because of using still the old deprecated bower-registry url.
I fixed it adding the ```.bowerrc``` file with the specific setting for registry ([ac8267c](https://github.com/selectize/selectize.js/pull/1426/commits/ac8267c333037ff7c92db29c7b53a475b7d86571)).

**Am not sure why we needed that:** on my machine (with all the same versions of npm, nvm, bower, etc than in the repository Travis-CI) I got no problems; it builds correctly.
As you can see at the linked Travis report, bower was failing to get dependencies, even if ```npm install bower``` should update to the latest version and get the new registry url.
Not sure why this happens.

Being unable to debug it I just opted for the ```.bowerrc``` fix, and **makes pull-requests pass again the checks correctly**, but maybe we want to get rid of that file and understand why in first place we get the issue and the wrong bower registry on travis-ci, when not specifying it manually.

Sorry for my usual verbosity, and ping me back for any concern 😉

Issue #1218, opened 2016-12-09
npm package should not have "dependencies", just "devDependencies"


vivainio:
We want to install the 'selectize' npm package just to get access to dist/selectize.js - i.e. we want to use the package similarly to the way you use it from bower (without browserify/webpack).

Currently package.json contains:
```
  "dependencies": {
    "jquery": "^1.7.0",
    "microplugin": "0.0.3",
    "sifter": "^0.5.1"
  }
```

We don't want to install any of these, and sifter is a problematic dependency because of the transient C++ dependency.

It would be useful to either have this changed, or publish a new npm package without production dependencies.

joallard:
In my understanding, dependencies are packages you need to run the library, devDependencies being packages you need to work on improving the library. Hence, we're saying Selectize cannot be ran without packages Jquery, Microplugin and Sifter.

Can you walk me through your logic?

vivainio:
Dependencies are the packages you actually require() in node sense, not packages you need to have loaded up by <script>. Selectize doesn't require() anything when included through selectize.js

For comparison, see e.g. https://github.com/angular-ui/bootstrap/blob/master/package.json - it has devDependencies to angular & others, but no actual "dependencies". Dependencies for frontend package don't make much sense unless you only support browserify or webpack workflows.

brianreavis:
When I set this up a long time ago, I included dist files for people manually downloading files into their project. That workflow has changed a lot by now. That said, I think including a package via a `<script>` tag loaded with NPM is not that common. (I could be proven wrong.) Most people are loading selectize via npm using it with a bundler like webpack or browserify. 

I don't forsee supporting separate packages for the myriad of build setups out there. It's up to @joallard though, as he's been the one graciously maintaining the project.

joallard:
Thanks for the input @brianreavis. This is not an unsignificant change, and I'm not well-versed enough in the different build environments out there to make that change. I'd be looking forward to hearing more perspectives about this though.

vivainio:
@brianreavis There has been an exodus of Bower users to NPM, and what they expect to do is just loading up their packages from node_modules instead of bower_components - still expecting a similar dependency-clean dist.js.

We have a good bunch of Angular frontend packages (fetched from npm) in our app, and none of them pulled runtime dependencies on installation.

heldchen:
[...]

PR #1430, opened 2018-07-17
[WIP] Move jQuery package.json dependencies

### LINKED ISSUE #1428: What's changed in 0.12.6
All I can see is version numbers being updated?  This version break support for jquery v1.  Thanks.

Pictor13:
and update deps and lock files.

- jQuery is a runtime dependency

closes #1428


graingert:
This will cause duplicate jQuery dependencie which isn't supported by jQuery

graingert:
Oh that might be better

Pictor13:
Yep, I was hitting my head on those WARNS until I didn't check the *npm semver* syntax 🙂

With selectize@0.12.6 it can't resolve a valid version:
https://asciinema.org/a/C2XsguyRsXQObmsWHbdH4MeK7

With the pull-request it gets rid of the WARN and satisfies the required dependencies:
https://asciinema.org/a/bT3xwHHaPCrHpVzUOR7YcOFwC (don't pay too much attention to the dyslexic typing...)

---

About the ```dependencies``` / ```devDependencies```, I still believe jQuery to be a runtime dependency.
Am not sure why as ```devDependencies``` it would not get duplicated (not sure about the little differences in how *dev* and *non-dev* are treated during install).

I couldn't manage to install and get a use case with duplicated versions. But maybe that's also depending on a different NPM version.
In which cases would we get the duplicated dependency that you mention?

graingert:
devDeps are ignored by upstream deps

Pictor13:
> devDeps are ignored by upstream deps

Yes, but
```devDependencies``` are for the dev-scripts, e.g. unit testing, packaging, doc generation
```dependencies``` are required for production use, and assumed required for dev as well.

When Selectize is the only project dependency requiring jQuery, no jQuery will be installed on *production*. There will be a PEER DEP warning (if the user notices that) but why forcing to act manually if there is no conflict?
Especially since our users didn't need to do it until now (= more people opening issues)

jQuery in ```dependencies``` will always work as expected from a non-expert user, and will warn just when an unsolvable conflict (duplicate) is found.
In that case the user MUST do something manually and investigate more.

I see most of the other libraries using jQuery to rely (imho correctly) on ```dependencies``` and ```peerDependencies```. I think we should do the same.

graingert:
jQuery is specifically mentioned in the npm docs as a case to use peerDeps

graingert:
Also npm won't warn on duplicate dependencies. It will just break in
production or result in duplicate jQuerys

Pictor13:
I also agree with the idea in [this article](https://codingwithspike.wordpress.com/2016/01/21/dealing-with-the-deprecation-of-peerdependencies-in-npm-3/), that the duplicates case is acceptable since it can be easily resolved, while maintaining compatibility with either NPM 2 & 3.

joallard:
Thanks for tackling this whole matter @Pictor13, I knew from the start it was going to be a thorny issue. I don't have any bandwidth to study this in depth, @Pictor13 @graingert do we have a consensus to merge?

graingert:
@joallard no. Do not merge this PR

joallard:
Thanks. Let me know when you're ready

Pictor13:
Don't merge yet.
I have the feeling that choosing the wrong direction here could provoke a lot of problem to old time users. And am not yet 100% convinced of either approach.
We should beware of as much use-cases and setups as possible and try to not create problems to them.

Anyway I add also #1218 to support @graingert point.
Still, @brianreavis answers to it push in the opposite direction. 🤔

joallard:
I really appreciate that we're taking the time to think about the consequences and that we're trying to reach a solution that makes some kind of consensus. If I can see that and that the changes are reasonable for most users going forward, I'm pretty sure I'll have @brianreavis' support on this.

My understanding of #1218 is that it's been more or less solved by moving jQuery to peerDependencies.

Edit: I put a WIP tag on the title, feel free to remove it when it's ready.

PR #1443, opened 2018-08-28
Added dropdown item activation events

michael-maltsev:
I use these for the following university scheduling project:
https://cheesefork.cf/
As you can see, when an item is activated in the dropdown, it's also previewed in the calendar.
I'm sure that it can have other uses.


archangel2080:
I could also use this.  Anything need to be done to get this merged in?

risadams (approved):

PR #1445, opened 2018-08-28
Fix measureString to not show a scollbar on IE/Edge, fixes #1444

### LINKED ISSUE #1444: Horizontal scrollbar on RTL pages in IE/Edge
I did:

	  like below

Steps to reproduce:

1. Visit https://selectize.github.io/selectize.js/ in IE/Edge.
2. Use inspector to set `dir='rtl'` for the `html` tag.
3. Observe the horizontal scrollbar which goes very far.

This happens because of the `measureString` utility, which creates a control which meant to be hidden far in the left:
https://github.com/selectize/selectize.js/blob/8c76832f95bae664fb1b533c41df414fc6192ebd/src/utils.js#L233-L260

michael-maltsev:
Fixes https://github.com/selectize/selectize.js/issues/1444

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -238,12 +238,17 @@ var measureString = function(str, $parent) {
 	if (!Selectize.$testInput) {
 		Selectize.$testInput = $('<span />').css({
 			position: 'absolute',
-			top: -99999,
-			left: -99999,
 			width: 'auto',
 			padding: 0,
 			whiteSpace: 'pre'
-		}).appendTo('body');
+		});
+
+		$('<div />').css({
+			position: 'absolute',
+			width: 0,
+			height: 0,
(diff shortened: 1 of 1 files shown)


Litecoinlord:
Fix measureString to not show a scollbar on IE/Edge and Firefox on An…

traviskroberts:
Selectize appending that test element to the body is causing accessibility issues (specifically, "Ensures all page content is contained by landmarks" - [more info](https://dequeuniversity.com/rules/axe/3.5/region)). Would it be possible to get the width of the temporary element, remove the element, then return the width?

The fact that this PR is almost two years old and hasn't received any attention makes me think it's probably never going to get merged though… ☹️

risadams (approved):

PR #1453, opened 2018-10-14
Fix "Invalid argument" in IE11

aarcangeli:
Fixes #1453

The problem occurs when the element is not a 'select'

Example:

    $('div.mySelect').selectize(options);


scipper:
when will this be merged? I have the same problem, when wrapping selectize in an AngularJs directive.

knoxcard:
merge merge merge!

PR #1460, opened 2018-11-16
A more performant updateOriginalInput

deefour:
See #910 for context.

This is a new PR rebasing the last 3 years of changes in `master`.

### DIFF
--- a/dist/js/selectize.js
+++ b/dist/js/selectize.js
@@ -2205,32 +2205,58 @@
 		 * Refreshes the original <select> or <input>
 		 * element to reflect the current state.
 		 */
-		updateOriginalInput: function(opts) {
-			var i, n, options, label, self = this;
-			opts = opts || {};
-	
-			if (self.tagType === TAG_SELECT) {
-				options = [];
-				for (i = 0, n = self.items.length; i < n; i++) {
-					label = self.options[self.items[i]][self.settings.labelField] || '';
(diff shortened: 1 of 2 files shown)


Pictor13:
I didn't review the code but the change looks interesting and useful. Good job 👍🏻

As you can there are errors with Travis CI.
In order to submit a pull request correctly, just the source code under `src/` directory should be modified, leaving untouched the code under `dist/` (distributed build).
This latter will be automatically updated with all the latest pull-requests, once there will be a new build release.

deefour:
I don't plan on putting any more work into this.

deefour:
It would be a shame for this PR to get auto-closed out by a Github action. While I cannot comment on a recent version of selectize, at the time the code in this PR was the difference between the dropdown box being unusable vs very performant.

As I mentioned 1.5 years ago, I won't be putting more work into this, but if it hasn't otherwise been fixed it'd be a shame not to not see the performance boost make it into a release.

risadams:
Update on this PR,  I haven't merged it yet because it is causing tests to fail; however I'm not particularly sure that the original code is working quite correctly, so until I can determine what the correct behavior is, this is going to sit in the pending queue for a bit

marianoesteban:
Escaping `value` and `label` with `escape_html()` in line 1742 makes the tests pass:
```js
fresh.push('<option value="' + escape_html(value) + '" selected="selected">' + escape_html(label) + '</option>');
```

risadams:
Rebased, updated and merged in new PR #1895

PR #1461, opened 2018-11-20
chore(dependencies): fix security vulnerabilities

joaogarin:
Hello,

I am providing here an update with security vulnerabilities in selectize fixed. All tests have passed and also I did a try on a site using selectize and all seems ok. So would be great to have this fixed since currently people that might (for whatever reason) include the repo dont get security warnings..also the repo itself would be in better shape in terms of security.

let me know if there is anything else I can do here.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,451 +4,312 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
-    "abbrev": {
-      "version": "0.12.6",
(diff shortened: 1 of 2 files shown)


klausi:
Thanks, this hopefully resolves our Github security warnings about package-lock.json which we include in a Drupal project.

joaogarin:
cc @joallard

PR #1474, opened 2019-02-11
Add any data-attributes to static select options

warloque:
Read static select and select options' data-attributes, so it should be possible to insert custom html-content to options, search for option by custom keywords and so on. 

At present, only value and label static option values are read on initialization.


knoxcard:
Sounds awesome!

Issue #2, opened 2013-05-28
click on scrollbar closes the suggestion-box


milworm:
http://brianreavis.github.io/selectize.js/ - the last example.

try to click on the scrollbar.

Issue #1, opened 2012-12-07
Add option for confirmation on item delete.


brianreavis:
Right now it's really easy to accidentally delete items... in some cases this is great (tagging), in other cases this is bad.

Issue #1492, opened 2019-06-12
[PR Available] Option group header isn't displayed if value is zero (0) or boolean


winzig:
I initiate Selectize using data, not a `<select>` element, and have encountered a bug with option group headers. 

I initialize the Selectize object with this:

```
optgroupField: "Type",
optgroups: [
    {value: 0, label: 'My Label for Zero'},
    {value: 1, label: 'My Label for One'},
    {value: 2, label: 'My Label for Two'},
    {value: 4, label: 'My Label for Four'}
],
```

Notice that "My Label for Zero" has a `value` of `0` (zero). 

Each of my options has a `Type` property equal to 0, 1, 2, or 4. However, the group represented by the `Type=0` never has its header printed in the dropdown. All the other ones appear normally. 

While stepping through the JS debugger where Selectize is working on rendering the option groups, I am seeing a clue. Internally, Selectize is representing the "zero" Type as `""` (empty string), instead of zero:

<img width="769" alt="Screen Shot 2019-06-12 at 10 09 24 AM" src="https://user-images.githubusercontent.com/127062/59371604-59ee1a80-8cfa-11e9-8ce6-60328105ca6a.png">

In that block you can see it's going to fail on the next line down when it reaches this logic:

`if (self.optgroups.hasOwnProperty(optgroup) ...`

Because optgroups has properties of (0, 1, 2, 4), and not `""`, which is what the value of `optgroup` is in this loop. 

I'll update this issue if I can narrow this down further.

winzig:
I'm pretty sure the issue is with this line:

https://github.com/selectize/selectize.js/blob/8c76832f95bae664fb1b533c41df414fc6192ebd/src/selectize.js#L1086

For my data, the value of `option[self.settings.optgroupField]` is 0, and so in that code, is interpreted as false, and thus the optgroup variable gets set to empty string, which is not part of my data.

winzig:
If I change the above referenced line to this:

```
optgroup    = option.hasOwnProperty(self.settings.optgroupField) ? option[self.settings.optgroupField] : '';
```

Then it works as expected. Instead of just checking to see if the value is truthy, it's checking just to see if it exists, and then if so, uses the value as-is. 🎉 

[...]

PR #1493, opened 2019-06-13
Allow 0 and false to be used in opt group values

winzig:
Refer to this discussion: https://github.com/selectize/selectize.js/issues/1492

### DIFF
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -1083,7 +1083,7 @@ $.extend(Selectize.prototype, {
 		for (i = 0; i < n; i++) {
 			option      = self.options[results.items[i].id];
 			option_html = self.render('option', option);
-			optgroup    = option[self.settings.optgroupField] || '';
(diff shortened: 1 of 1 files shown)


winzig:
Long story short: When you define optgroups, you specify values for Selectize to match a certain field against, from your options data. It uses that field to group your options together and display headers for each option group. However, there is a bug (which this fixes) whereby if your values are falsy (like zero `0` or `false`, things don't work. 

There are two bugs in the line that is fixed: 

1. It was doing an or `||` test to determine if your value exists, and if your value is falsy, it's turned into empty string, which won't match your defined values. 
2. Even when the bug in #1 is fixed, elsewhere in the code, the values are getting passed through the `hash_key()` function to ensure all the values get turned into strings. In this case, hash_key() turns false into '0' (zero string). Yet in this section of code that I'm fixing, the compared value was not passed through the hash_key() function, so it was comparing apples to oranges. 

To fix #1, we check for existence using `hasOwnProperty()` instead of a simple boolean test, and to fix #2 we pass this value through `hash_key()` so that it matches how the values are manipulated elsewhere in Selectize.

Issue #1506, opened 2019-10-22
Vulnerability in csv-parse version in yarn.lock


iHiD:
Hello, and thank you for your work with selectize :)

A security vulnerability has been announced for csv-parse - [CVE-2019-17592](https://nvd.nist.gov/vuln/detail/CVE-2019-17592).

According to the GitHub security advisory, the remediation is:
> Upgrade csv-parse to version 4.4.6 or later. For example:

```
csv-parse@^4.4.6:
  version "4.4.6"
```

We are currently blocked from updating because selectize relies on an older version of `csv-parse`. Could you consider updating `csv-parse` in this project please? If that's not possible, is there a way for us to update `csv-parse` and use `selectize`

Thanks!

mattgodbolt:
The `sifter` package needs to be updated first: https://github.com/brianreavis/sifter.js/pull/56 tracks this.

iHiD:
The dangled web of dependencies!

Thanks @mattgodbolt :)

alec-joy:
@brianreavis hasn't commented on the open issues or PRs about this in almost 2 months, it might be time to consider a fork for this dependency. I understand @brianreavis is also one of the primary authors of this project so if there's any change someone could get in touch with him about this to avoid a fork I'm all for it. I totally understand if they don't have time to deal with this, but some resolution would be very helpful in the near future.

Xenology:
His profile is saying that just today he has had commits, so I'm hopeful some tagging might get his attention. He's got a personal website in his bio with an email. I'd like to get this resolved before the end of the year on my end, as I'm getting really tired of Github's automated messages around vulnerabilities. If we don't hear from him in the next few weeks, it might be good to reach out over email. Maybe he'd just be willing to hand over the repo to someone else to maintain.

mattgodbolt:
Thanks @alec-joy and @Xenology . And also ping @brianreavis :) pretty please :)

leandrocrs:
I've sent a email message and a tweet to @brianreavis remembering him that PRs are already open.
I hope do not bother him 🙃 

https://twitter.com/brianreavis

brianreavis:
Please see https://github.com/brianreavis/sifter.js/pull/56#issuecomment-572253973

PR #1519, opened 2020-01-31
Update Sifter to fix security vulnerabilities

naganowl:
Mainly to update `csv-parse` which has had a CVE issued for it.

Addresses https://github.com/selectize/selectize.js/issues/1506 and follows the pattern from the previous [dependency update](https://github.com/selectize/selectize.js/pull/1461).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "integrity": "sha1-63d99gEXI6OxTopywIBcjoZ0a9I=",
       "dev": true,
       "requires": {
-        "mime-types": "~2.1.18",
+        "mime-types": "2.1.21",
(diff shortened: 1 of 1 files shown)


Pictor13:
Thanks!

Unfortunately it has conflict with the `package-lock.json`.
**EDIT:** oh, I see, it may be because @risadams force-pushed ?
Maybe you need to `git pull --rebase` and reapply the fix?

However, the last comment of the issue you linked points to [another issue, on Sifter repo](https://github.com/brianreavis/sifter.js/pull/56#issuecomment-572253973), explaining that Selectize will not incur in this security issue.

risadams:
sifter has been updated to the latest published version in project root.

PR #1520, opened 2020-02-05
Update selectize.js

ymeric:
isFull function is returning true although create option is false. This blocks search feature in certain scenarios. This change will make sure the create option is true before comparing maxItems to length of items.


Pictor13:
I remember some bug like that....
I've never had the time to do a pull-request; thanks!

I think adding a test case to ensure it works would be great.

Looks like there's a new maintainer now, so PR are being evaluated again :) 
But the PR should pass the tests, I guess.

PR #1528, opened 2020-04-23
Fix mouse blur events

davidmason:
I wasn't getting focusout events when clicking out of selectize elements with mouse. Keyboard navigation did work.

Implementing the fix suggested here: https://github.com/selectize/selectize.js/issues/406#issuecomment-59204127


davidmason:
This doesn't seem to fix the issue I was having, so better to leave it as-is.

PR #1536, opened 2020-06-11
onClick opens the dropdown

bharatsamurai:
Added "self.open();" in onClick() event at line 1497 to open the dropdown list with single click inside the input field. Earlier one needs to click multiple times or navigate with the down arrow key only.

### DIFF
--- a/dist/js/standalone/selectize.js
+++ b/dist/js/standalone/selectize.js
@@ -1494,6 +1494,7 @@
 	    // closeAfterSelect
 			if (!self.isFocused || !self.isOpen) {
 				self.focus();
+				self.open();
 				e.preventDefault();
 			}
 		},


Pictor13:
Thanks. Unfortunately the pull request is invalid:
you changed the distribution compiled code; while you should have changed it under the `src/` folder (that later on is automatically compiled to the built version under `dist/`).
I made the same mistake at my first Pull-Request on Selectize 😉

However, I am not sure that it is the right fix.
From what I remember there was some setting (`openOnFocus`?) to decide if the dropdown should be opened or not, at `onFocus`.
IIRC is the `refreshOptions` function that decides what to do with the dropdown, based on the `triggerDropdown` flag.

https://github.com/selectize/selectize.js/blob/0c5c6173bbd2f6bafd8f4baccf0c91106d4fbab1/dist/js/standalone/selectize.js#L1747
https://github.com/selectize/selectize.js/blob/0c5c6173bbd2f6bafd8f4baccf0c91106d4fbab1/dist/js/standalone/selectize.js#L2306

If you want to fix the pull-request I would ensure that:
- `npm test` runs correctly, on you development machine.
- the change is really in the correct point **AND** works in *all cases* (that are a lot to check without relying on tests: multiple, single, multiple+maxItems:1, and others).

PR #1546, opened 2020-09-17
docs: Fix simple typo, hightlight -> highlight

timgates42:
There is a small typo in dist/js/selectize.js, dist/js/standalone/selectize.js, src/contrib/highlight.js.

Should read `highlight` rather than `hightlight`.


risadams:
This will be fixed,  good catch.
this PR should not have included files within the dist folder, as those will get regenerated.

PR #1576, opened 2021-01-17
Autoselect option based on typed text on focusout

anandphulwani:
- Added onBeforeBlur functionality.
   - On a blur event, before calling **onBlur**, an **onBeforeBlur** function is called, which checks that the text typed in the input box is an exact match of any option provided in the select control ( letter case is ignored ), if an option is found, that option is selected.
-  Added functions like getTextboxValue, getElementWithTextContent, getFirstItemMatchedByTextContent
   - **getTextboxValue**: Gets the current value of input control.
   - **getElementWithTextContent**: Finds the first element with a "textContent" property (innerHTML) that matches the given string provided for search..
   - **getFirstItemMatchedByTextContent**: Returns the jQuery element of the item matching the given textContent / innerHTML.

### DIFF
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -762,6 +762,16 @@ $.extend(Selectize.prototype, {
 	},
 
 	/**
+	 * Gets the value of input field of the control.
+	 *
+	 * @returns {string} value
+	 */
+	getTextboxValue: function() {
+		var $input = this.$control_input;
+		return $input.val();
+	},
+
+	/**
 	 * Sets the input field of the control to the specified value.
 	 *
 	 * @param {string} value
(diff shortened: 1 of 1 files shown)


src/selectize.js
```diff
@@ -1466,6 +1516,18 @@ $.extend(Selectize.prototype, {
...
 
+	/**
+	 * Returns the jQuery element of the item
+	 * matching the given textContent.
+	 *
+	 * @param {string} value
+	 * @param {boolean} ignoreCase
+	 * @returns {object}
+	 */
+	getFirstItemMatchedByTextContent: function(textContent, ignoreCase = false) {
```
risadams: remove  '= false'  from the ignoreCase argument, and instead check and initialize the default inside the method body

risadams (changes requested):

anandphulwani:
Reworked the code and sent a different PR for the same.

PR #1577, opened 2021-01-17
fix bootstrap4 theme

olegatro:
If we have situation `<select class="form-control form-control-sm multiple"></select>` we get bad result because of fixed `height` with `important`.


risadams (approved):
Seems straightforward,  I'll run a quick visual test and merge when I can.  Thanks.

olegatro:
@risadams  you should check in the multiline situation. 

example https://codepen.io/olegatro/pen/ExgrKEQ

Issue #996, opened 2016-01-28
UX improvement: select all text in input on focus


nugged:
when I press on drop-down triangle I have list and cursor blinking in the input field:
![image](https://cloud.githubusercontent.com/assets/295324/12661060/2408b21c-c620-11e5-896e-efbe65e0feef.png)
And yes, it that field definitely one whole item from the list, so I can't input any more letters but must press backspace, it erases all text, and I start to type new with pretty well-designed typeahead search.
Well.

But what if we will select all text in input on focus, so on _any_ user input it will start a search from scratch, I mean no need to press backspace and just any first alphanumeric character pressed start to search new item?

Or if I just missed that option from manual, please, help/advice.
Or should I make it with a snippet, what should be there for 'focus' event below?
(how to get that `$(this).{...somethow_get_to_that_internal_input_field...}.select();` working?)

```
$('#forminput_tz').selectize();
selectize.on('focus', function() {
    ... what should be here?
    $(this).{...somethow_get_to_that_internal_input_field...}.select(); ?
});
```

thedustyrover:
+1 on this, this was my default UX expectation. With the current impl I think it would be hard to do as the label actually exists not within the input element, but a separate div. If selectize could populate the value directly into the text input and gave an option to select() whenever it gained focus, that would fulfill this request I think. But I imagine that it would be a breaking change to get rid of the div.

As a partial solution, I've done the following to add a css highlight to the div when the input has focus:

```
.selectize-input.input-active > .item {
    background:#CCC;
}
```

it gives the appearance that it's selected and an affordance that the user can backspace to remove it. The two downsides are:
1. difficult if not impossible to match the actual text-selection color, since that's dependent on browser/OS.
2. it's still only backspace that works, you cannot immediately start typing over it.

thedustyrover:
I started on a crappy hack to get the "immediately start overwriting if a character is typed rather than requiring a backspace." I abandoned it because I didn't want to get into all the nitty-gritty of implementing the advanced keystroke checks, but here's a pointer if anyone else wants to go down that route:

```
$('.myInput')[0].selectize.$control_input.keydown(function(e) {
    var selectize = $('.myInput')[0].selectize;
    // if there exists a selection already and the text input is empty
    // otherwise if the selection is empty, the text input is already working like normal
    // same with non-empty text input, we don't want to interfere with regular typing
    if(selectize.items.length != 0 && selectize.$control_input.val() == '') {
        selectize.clear();
[...]

PR #1578, opened 2021-01-18
Selects all text on focus

anandphulwani:
- Selects text on focusing into the element by keyboard.
- Also highlights the option in the dropdown.

This should resolve the issue mentioned here https://github.com/selectize/selectize.js/issues/996

To use just do 
```
element.selectize({
    plugins: ['selecttext_on_focus']
});
```

### DIFF
--- a/src/plugins/selecttext_on_focus/plugin.js
+++ b/src/plugins/selecttext_on_focus/plugin.js
@@ -1,0 +1,54 @@
+Selectize.define('selecttext_on_focus', function(options) {
+	var self = this;
+	
+	var currValue;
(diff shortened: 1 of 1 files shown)


anandphulwani:
This will work only when this PR https://github.com/selectize/selectize.js/pull/1576 is merged first.

anandphulwani:
This plugin runs great with 0.12.6 but is giving a recursive error with 0.13.0, put a hold to this before i sort this out.

anandphulwani:
Reworked the code and sent a different PR for the same.

PR #1583, opened 2021-01-20
add browserslist file

olegatro:
preparation for add autoprefixer tool


risadams (approved):
I will merge this, but update the supported list to be more modern

PR #1588, opened 2021-02-06
Don't replace all versions in package-lock.json during release

php-coder:
This change doesn't fully fix the original issue but it reduces impact area as now we only replace old version. The better solution is to replace only the first occurence or replace a version only if it is the next line after a package name (this can be done wih multiline regexps).

P.S. This problem took my attention only because you wrote a good commit message  in 0979580bf224151e4fa3db1b217e4e2e87ce99f1 ;)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 .PHONY: compile release test
 plugins=*
 GRUNT=node_modules/.bin/grunt
+CURRENT_VERSION := $(shell sed -n '/"version":/{s/.*"version": "\([^"]*\)".*/\1/p;q}' package.json)
 
 all: compile
 test:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,6 +1,7 @@
 .PHONY: compile release test
 plugins=*
 GRUNT=node_modules/.bin/grunt
+CURRENT_VERSION := $(shell sed -n '/"version":/{s/.*"version": "\([^"]*\)".*/\1/p;q}' package.json)
```
php-coder, line 4: Here is how this sed expression might look like in a more human-readable form:
```sed
# if we found a line that contains this pattern
/"version":/ {
    # extract a version out of this line and print it
    # (otherwise it won't because we passed `-n` option)
    s/.*"version": "\([^"]*\)".*/\1/p;
    # stop script execution:
    # we don't want to check other lines as we need only the first occurence  
    q
}
```

php-coder: P.S. Hm... but this only work with GNU sed. Should we care about it?

risadams: Not sure we need to care, as this isn't something that is run often, and we usually have control over the environments.  If we need to do something more robust later we can.

Good Work

risadams (approved):

PR #1589, opened 2021-02-07
Updated usage.md

anandphulwani:
Changed usage.md to show remaining columns in github.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -418,23 +418,23 @@ $(function() {
 			<table width="100%">
 				<tr>
 					<td valign="top"><code>option</code></td>
-					<td valign="top">An option in the dropdown list of available options.</td>
+					<td valign="top">An option in the <br />dropdown list of <br />available options.</td>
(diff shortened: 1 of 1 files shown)


risadams (commented):
Unless I'm missing something I don't see any changes other than adding line breaks; I'm not quite sure what value this is providing.

anandphulwani:
Actually it is for better readability, github doesn't have wrapping feature in the inner tables, and i really didn't know for quite long that there are other columns as well apart from the first two.
This is how the current usage.md looks.

recording

and you can have a look at what i have modified over here
https://github.com/anandphulwani/selectize.js/blob/Update-usage.md-to-show-columns/docs/usage.md

anandphulwani:
@risadams 
> and you can have a look at what i have modified over here
> https://github.com/anandphulwani/selectize.js/blob/Update-usage.md-to-show-columns/docs/usage.md

Please check this as well.

PR #1590, opened 2021-02-07
Resolved focus cycle problem in firefox.

anandphulwani:
For those dropdown where it doesn't show scrollbar the focus cycle works properly, but for those dropdowns having enough options to show scrollbar the focus cycle gives problem.
This commit removes the scrollbar div from the focus group by adding tabindex="-1" to the dropdown.


risadams (approved):
I'll tweak this a bit to be more consistent with surrounding code,  but good find,

anandphulwani:
Do keep me in the loop, i want to know how can i improve more.

risadams:
it wasn't much of a change,  just moving the tabindex into an attr() call to match the rest of the code
https://github.com/selectize/selectize.js/commit/1fe1df339c74e6c5f079e4ffb661ad03e18055e6

PR #1595, opened 2021-02-23
Resets the dropdown selection to top, if input is cleared.

anandphulwani:
When user types in the input field, and a particular option gets selected, if the user clears the input field, the older option remains selected, this PR rectifies by selecting the topmost option.
### **Older:**
older

### **Post PR:**
Once this patch is applied you can see post clearing the input field,
the selection moves to the top most element.
post_pr


risadams (approved):
Nice change 👍

PR #1599, opened 2021-02-24
Update backspace action on dropdown.

anandphulwani:
Whenever we clear an option using backspace, the selection moves one step upward. This creates confusion as when user deletes and presses tab (given `selectOnTab: true` ), the above option gets selected automatically, whereas deletion should move it to the top of the list.

### Before PR
Before

### After PR
After

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -148,6 +148,12 @@ $(function() {
 		<td valign="top"><code>60</code></td>
 	</tr>
 	<tr>
+		<td valign="top"><code>dropdownOnBackspaceGotoTop</code></td>
+		<td valign="top">If an option is selected, the same option is highlighted/marked active in the dropdown, pressing backspace on the input control removes the option and in dropdown the previous element is highlight. When this option is set to `true` it shifts the highlight to the topmost option.</td>
+		<td valign="top"><code>boolean</code></td>
+		<td valign="top"><code>false</code></td>
+	</tr>
+	<tr>
 		<td valign="top"><code>loadThrottle</code></td>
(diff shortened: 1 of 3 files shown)


risadams:
Can this be turned into a preference so that it can be a choice to go to the top of the list or the previous option?

anandphulwani:
@risadams Can be done, i will look into it and update it.

anandphulwani:
@risadams Done. Please check.

src/defaults.js
```diff
@@ -23,6 +23,7 @@ Selectize.defaults = {
 	closeAfterSelect: false,
 
 	scrollDuration: 60,
+	dropdownOnBackspaceGotoTop: false,
```
risadams: the name is a bit awkward.
Maybe something like: deselectBehavior? 
Maybe instead of a boolean we can make it a string/enum to allow for other options in the future?

docs/usage.md
```diff
@@ -147,6 +147,12 @@ $(function() {
 		<td valign="top"><code>int</code></td>
 		<td valign="top"><code>60</code></td>
 	</tr>
```
risadams, line 149: see comment in src/deafaults.js
If that is updated this would need updated too

risadams (changes requested):

src/defaults.js
anandphulwani: Can be done, will do, and get back to you.

docs/usage.md
anandphulwani: Noted.

anandphulwani:
All changes done.

risadams (approved):

PR #1603, opened 2021-02-25
Show empty option in dropdown and label for that empty option.

anandphulwani:
Added the following settings.
### 1. showEmptyOptionInDropdown ( Default: false )
If true, Selectize will show an option with value `""` in dropdown, if one does not exist; which is useful when you want to select a empty option via `selectOnTab`. This requires `allowEmptyOption` also set to `true`.

### 2. emptyOptionLabel ( Default: -- )
If `showEmptyOptionInDropdown: true` and an option with value `""` in dropdown does not exist, an option with `""` value is created, the label/text of the option can be set via this option, this requires `showEmptyOptionInDropdown: true`.

### 3. Added class `selectize-dropdown-emptyoptionlabel` to center align the `emptyOptionLabel`

Post PR
after

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -142,6 +142,18 @@ $(function() {
 		<td valign="top"><code>false</code></td>
 	</tr>
 	<tr>
+		<td valign="top"><code>showEmptyOptionInDropdown</code></td>
+		<td valign="top">If true, Selectize will show an option with value `""` in dropdown, if one does not exist; which is required when you want to select a empty option via `selectOnTab`. This requires `allowEmptyOption: true`.</td>
+		<td valign="top"><code>boolean</code></td>
+		<td valign="top"><code>false</code></td>
+	</tr>
+	<tr>
+		<td valign="top"><code>emptyOptionLabel</code></td>
+		<td valign="top">If `showEmptyOptionInDropdown: true` and an option with value `""` in dropdown does not exist, an option with `""` value is created, the label/text of the option can be set via this option, this requires `showEmptyOptionInDropdown: true`.</td>
+		<td valign="top"><code>string</code></td>
+		<td valign="top"><code>'--'</code></td>
+	</tr>
+	<tr>
 		<td valign="top"><code>scrollDuration</code></td>
 		<td valign="top">The animation duration (in milliseconds) of the scroll animation triggered when going [up] and [down] in the options dropdown.</td>
 		<td valign="top"><code>int</code></td>
--- a/src/defaults.js
+++ b/src/defaults.js
@@ -20,6 +20,8 @@ Selectize.defaults = {
 	selectOnTab: true,
 	preload: false,
 	allowEmptyOption: false,
+	showEmptyOptionInDropdown: false,
+	emptyOptionLabel: '--',
 	closeAfterSelect: false,
 
 	scrollDuration: 60,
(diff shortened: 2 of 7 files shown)


risadams (approved):

jicihome:
Actually It doesn't seem to work for me. Do we need to load options in a specific way? (Example: will only work if data are loaded from options field and not from an array like [{"display":"test to display","value":"12345"}] ?

anandphulwani:
Can you upload the snippet of your code, what are you using, what part is working for you and what is not?

As far as i understand from your question, data can be loaded in selectize from `<option>`  tag under `<select>` and also by passing array in the JS constructor, in the options key.
> 
> Actually It doesn't seem to work for me. Do we need to load options in a specific way? (Example: will only work if data are loaded from options field and not from an array like [{"display":"test to display","value":"12345"}] ?

jicihome:
I mean the part that showEmptyOptionInDropdown: true, from what i understand, should add an empty option in the list if there's none. It doesn't seem to work. Tried with the Basic page with allow empty by removing the empty options and add this settings and   no success.

				$('#select-beast-empty').selectize({
					allowEmptyOption: true,
					showEmptyOptionInDropdown: true,
					create: true
				});
					<select id="select-beast-empty" class="demo-default" data-placeholder="Select a person...">
						<option value="4">Thomas Edison</option>
						<option value="1">Nikola</option>
						<option value="3">Nikola Tesla</option>
						<option value="5">Arnold Schwarzenegger</option>
					</select>

anandphulwani:
Your code is perfect, check at the below codepen
The only problem is that we haven't compiled the code and created the dist js as of yet, as we are making rapid changes in the code. The next dist should be in 15 days and will contain a lot of new options, you are free to copy the js from my codepen and use it in the meantime.
[https://codepen.io/anandphulwani/pen/LYbvRde](https://codepen.io/anandphulwani/pen/LYbvRde)

jicihome:
Haa! Thanks a lot. Didn’t know. I was just too fast. Thanks again!

jicihome:
Is the new dist contain this update? @anandphulwani 
Thanks

jicihome:
Hi! Is there a plan to add this to standalone? thanks

anandphulwani:
@jicihome This shows as merged and i can see the following in https://github.com/selectize/selectize.js/blob/master/dist/js/selectize.js
```
(Line 2835)		showEmptyOptionInDropdown: false,
(Line 2836)		emptyOptionLabel: '--',
```

jicihome:
@anandphulwani This is not the standalone version. I'm asking about Standalone version.
Thanks

jicihome:
Hi @anandphulwani 
I get it working for a short time, but now it doesn't. Seem that the issue is related to how data are loaded. your codepen was using "select" but it doesn't seem to work when using options
https://codepen.io/jicihome/pen/GRQGqmQ

anandphulwani:
> Hi @anandphulwani I get it working for a short time, but now it doesn't. Seem that the issue is related to how data are loaded. your codepen was using "select" but it doesn't seem to work when using options https://codepen.io/jicihome/pen/GRQGqmQ

Hey, i am bit occupied for the next 3 months for some reasons, didn't want to keep you waiting for a reply, for the time being you are on your own. Hope you understand.

Issue #1156, opened 2016-09-23
single selectize control can't capture Chinese input string from virtual keypad on mobile device(IOS & Android)


thlo7777:
Steps to reproduce:
1. Change language to Chinese in iPhone or Android mobile 
2. Open a custom html page with select element, invoke selectize() options with create is true  in js file
3. Input Chinese characters to search or add new option

Expected result:
Same like English character, if options contains input character, then score and display, if not then show Add xxx .... 

Actual result:
Nothing happened, can't capture single Chinese character. 

[But I found when you input multiple Chinese characters, then use virtual keypad delete key, then the these characters will be captured. ]

Platform:  
iPhone 6 and Android + browser (webkit view)  + h5/js/css + bootstrap3.3.7 + selectize 0.12.3

I create a select 

`                    <div class="form-group">
                        <label for="cb-first-category" class="col-lg-2 control-label">cb商品一级分类</label>

```
                    <div class="col-lg-10">
                        <select class="form-control" id="cb-first-category" placeholder="选择/输入分类">
                            <option></option>
                            <option value="PA">Pennsylvania</option>
                              <option value="CT">Connecticut</option>
                              <option value="NY">New York</option>
                              <option value="MD">Maryland</option>
                              <option value="VA">Virginia</option>
                        </select>
                    </div>
                </div>
```

`  

When I input Chinese characters on mobile device (use standard iPhone virtual keypad and Chinese input method), the $control_input's  onKeyUp function can not capture it, but onKeyUp can get all English characters. 

I referenced the select2.js code (version 4.0.3)   [https://github.com/select2/select2]. They reminded a workaround to solve this issue.  (select.js  4.0.3  line 3918)
1. unbind keyup event from $control_input
2. bind ('keyup input') for $control_input.

Here are my codes, I tested and it works now.

`       var $select = $('#cb-first-category').selectize({
            create: true,
            sortField: 'text'
        });

```
    var selectize = $select[0].selectize;
    $('#cb-first-category-selectized').on('input', function(event) {
        $(this).off('keyup');
    });

    $('#cb-first-category-selectized').on('keyup input', function(event) {

        var self = selectize;

        if (self.isLocked) return e && e.preventDefault();
        var value = self.$control_input.val() || '';
        if (self.lastValue !== value) {
            self.lastValue = value;
            self.onSearchChange(value);
            self.refreshOptions();
            self.trigger('type', value);
        }
    });
```

`
[...]

PR #1608, opened 2021-03-18
Change 'keyup' event to 'input' event.

LI-NA:
Change keyup event to input event that used on detect changes of input (for search), because keyup event does not detect some input method (like mobile multi language keyboard, linux multi language IME).

Also deleted preventDefault method because both keyup and input event can not be canceled.

The input event was recommaned on MDN ( https://developer.mozilla.org/en-US/docs/Web/API/Document/keyup_event ).

Test passed: phantomjs, Firefox, Chrome. (Linux)
Related issues: #1156

* Sry for new test, I don't know how can I simulate multi language IME on phantomjs. :(

### DIFF
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -202,8 +202,8 @@ $.extend(Selectize.prototype, {
 		$control_input.on({
 			mousedown : function(e) { e.stopPropagation(); },
 			keydown   : function() { return self.onKeyDown.apply(self, arguments); },
-			keyup     : function() { return self.onKeyUp.apply(self, arguments); },
 			keypress  : function() { return self.onKeyPress.apply(self, arguments); },
+			input     : function() { return self.onInput.apply(self, arguments); },
 			resize    : function() { self.positionDropdown.apply(self, []); },
 			blur      : function() { return self.onBlur.apply(self, arguments); },
 			focus     : function() { self.ignoreBlur = false; return self.onFocus.apply(self, arguments); },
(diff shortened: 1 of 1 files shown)


demon101:
Some of my users see issue with copy-past to selectize without reaction. Never happens with old version. 
But I can't reproduce it in my browsers.

LI-NA:
> Some of my users see issue with copy-past to selectize without reaction. Never happens with old version.
> But I can't reproduce it in my browsers.

In theory, the issue does not happen if using input event, because input event will capture **every** changes of value on input element.

However, you mean the issue was occurred with current version (using keyup)? Keyup event cannot capture paste event without keyboard, so you can try paste text using mouse.

eduardo-mior:
Is there any forecast for this to be accepted?
I did some testing and onKeyUp actually has some issues that are fixed with onInput.
 
@LI-NA the utils.js file has a function called autoGrow() that uses the onKeyUp event, wouldn't it make sense to switch to onInput there too?

LI-NA:
@eduardo-mior Thanks for your advice. I didn't know there was a keyup event in utils.js, and I think it would works better if use an input event. (But it doesn't seem to be an issue even if we don't change it, because autoGrow use keydown and update events too.)

I believe there are no issues with this request, but I don't know why it's not approved yet... I'm sure input is better than keyup.

Here is small gif to test keyup vs input. Keyup cannot detect some IME.
input_test

eduardo-mior:
@LI-NA If you paste a text with the mouse onKeyUp is also not triggered (copy a text, and right-click on the input and click paste).
This is just another example of a bug that exists, on mobile this also happens, hold pressed and click paste.

eduardo-mior:
Any news? Merge?

risadams (approved):

PR #1754, opened 2022-02-05
Feature/dont fire change if nothing changed

rotdrop:
Don't fire a change event if setValue() did not change anything.


risadams (changes requested):
Can you please add a unit test for this?

PR #1756, opened 2022-02-05
refreshItems(): add "silent" option.

rotdrop:
This is meant for the case where refreshItems() is called
programmatically and hence the change event is not needed, or even bound
to a handler which rather would react to "external" user input.


risadams (changes requested):
Can you please add a unit test for this?

PR #1757, opened 2022-02-05
Open drop-down on mouse-click if openOnFocus is false and when clicking on the options.

rotdrop:
Otherwise: how can the drop-down be opened if openOnFocus is false?


fabienwnklr:
@risadams Hi, any reason for don't merge this PR ?

risadams:
rebased and merged in #1899

PR #1772, opened 2022-03-09
Update CONTRIBUTING.md

ghost:


risadams (changes requested):
While I agree that there needs to be something more here,  this isn't it :)

PR #1796, opened 2022-04-21
Enable setting the first option in the list as active.

Joshuan92:


famoser:
When the user types into the search bar, the first item is always highlighted. When the user then uses backspace (for example, to correct a typo), the same item stays highlighted even though it may not be the top-most item anymore. This is arguably confusing UX.

For example, the user types "Anna", highlighting the person named Anna. Then the user uses backspace to correct a typo. When the user pressed backspace twice (resulting in "An" as a search query), Anna remains highlighted even though there could be many more persons before (for example "Anastasia"). 

This option opts-in to always highlight the first element in the search results, independent of whether the user used backspace or not.

src/selectize.js
```diff
@@ -1200,23 +1200,25 @@ $.extend(Selectize.prototype, {
...
-				}
-				if (!$active || !$active.length) {
-					if ($create && !self.settings.addPrecedence) {
-						$active = self.getAdjacentOption($create, 1);
-					} else {
-						$active = $dropdown_content.find('[data-selectable]:first');
-					}
-				}
+    // activate
+    self.hasOptions = results.items.length > 0 || ( has_create_option && self.settings.showAddOptionOnCreate ) || self.settings.setFirstOptionActive;
```
risadams, line 1204: TODO: We should look at a cleanup of these if checks into something that is both easier to understand as a dev, and avoids edge cases

CHORE (testing, documentation):  Do you have any updates to unit tests or an example that can show the expected behavior?

famoser:
Could the maintainers please indicate whether they intend to review/merge this, or should I close this PR? This is still a concern from my point of view. Thanks!

risadams (approved):
dist files should not be modified here.  they will be reset during a build

risadams:
copied and merged in #1874

famoser:
Thank you a lot @risadams !

Issue #1689, opened 2021-10-25
Selectize dropdown selection outside of bootstrap modal window causes the modal to close


NashBridgesAT:
I did:

* [x] Search for if my issue has already been submitted
* [ ] Make sure I'm reporting something precise that needs to be fixed
* [x] Give my issue a descriptive and concise title
* [ ] Create a *minimal* working example on JsFiddle or Codepen
	(or gave a link to a demo on the Selectize docs)
* [ ] Indicate *precise* steps to reproduce in *numbers* and the result,
	  like below

I am using seltectize within a modal window. The problem is, that the dropdown extend beyond the bottom of the modal. Clicking an option beyond the bottom of the modal causes the modal to close. Only solution i found so far is to make the modal static, but that is not the behaviour i want. Is there any possibility to prevent the modal click event from being fired? Or any other solution for this issue?

Thanks a lot.

vanboom:
Wanting to contribute more detail to this issue.  Please consider reopening unless this may be a Firefox bug? 

Here is a test file that reproduces this behavior on Firefox.  Notice - clicking on a select option that drops outside of the modal will close the modal as Firefox propagates the click event.  Chrome works as expected.   Notice in the test, that a normal non-selectize select does not close the window.  I will dig deeper to see if I can isolate this and submit a PR if possible. 


```html

<!DOCTYPE html>
<html>
<head>
<script src='https://wwwqmx-13d4e.kxcdn.com/js/jquery-3.2.1.min.js'></script>
<script src='https://cdnjs.cloudflare.com/ajax/libs/microplugin/0.0.3/microplugin.min.js'></script>
<script src='https://cdnjs.cloudflare.com/ajax/libs/jquery-modal/0.9.2/jquery.modal.min.js'></script>
<script src='https://cdnjs.cloudflare.com/ajax/libs/selectize.js/0.13.5/js/selectize.min.js'></script>
<script src='https://cdnjs.cloudflare.com/ajax/libs/sifter/0.5.4/sifter.min.js'></script>
<link href='https://cdnjs.cloudflare.com/ajax/libs/jquery-modal/0.9.2/jquery.modal.min.css' rel='stylesheet'>
<link href='https://cdnjs.cloudflare.com/ajax/libs/selectize.js/0.13.5/css/selectize.css' rel='stylesheet'>
<link href='https://cdnjs.cloudflare.com/ajax/libs/selectize.js/0.13.5/css/selectize.default.css' rel='stylesheet'>
[...]

PR #1813, opened 2022-06-06
Issue #1689, change selectize behavior to work inside modals.

vanboom:
Thank you for selectize.js!  I am pleased to contribute this PR for your consideration.  

The root cause of issue #1689 is that the drop down is being closed on the `mousedown` action, leaving the `mouseup`/`click` event to happen outside of the control.  That causes the `blur` event which closes any modal that the selectize resides within. 

A standard HTML select control changes when the `mouseup` action is received.  This PR emulates that same behavior by making the selection change when the `mouseup` is detected. 

A logic error was also discovered where it appears that the author is attempting not to blur the control if the user clicks on a dropdown item.  See the code comments for discussion on this. 

Finally, IE11 should no longer be supported, so a few lines were removed that were causing further issues with the drop down flashing open then closed when a click was performed outside of the control. 

Thank you for your consideration.  I have tested this thoroughly on Chrome, Firefox, Safari on Windows, Linux and various tablet/mobile devices and it appears to be a significant improvement to the behavior of the control.

### DIFF
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -135,7 +135,7 @@ $.extend(Selectize.prototype, {
 
 		$wrapper          = $('<div>').addClass(settings.wrapperClass).addClass(classes).addClass(inputMode);
 		$control          = $('<div>').addClass(settings.inputClass).addClass('items').appendTo($wrapper);
-		$control_input    = $('<input type="text" autocomplete="new-password" autofill="no" />').appendTo($control).attr('tabindex', $input.is(':disabled') ? '-1' : self.tabIndex);
+		$control_input    = $('<input type="select-one" autocomplete="new-password" autofill="no" />').appendTo($control).attr('tabindex', $input.is(':disabled') ? '-1' : self.tabIndex);
 		$dropdown_parent  = $(settings.dropdownParent || $wrapper);
 		$dropdown         = $('<div>').addClass(settings.dropdownClass).addClass(inputMode).hide().appendTo($dropdown_parent);
 		$dropdown_content = $('<div>').addClass(settings.dropdownContentClass).attr('tabindex', '-1').appendTo($dropdown);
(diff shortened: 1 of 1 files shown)


src/selectize.js
```diff
@@ -190,10 +190,10 @@ $.extend(Selectize.prototype, {
 		self.$dropdown         = $dropdown;
 		self.$dropdown_content = $dropdown_content;
 
-		$dropdown.on('mouseenter mousedown click', '[data-disabled]>[data-selectable]', function(e) { e.stopImmediatePropagation(); });
+		$dropdown.on('mouseenter mousedown mouseup click', '[data-disabled]>[data-selectable]', function(e) { e.stopImmediatePropagation(); });
 		$dropdown.on('mouseenter', '[data-selectable]', function() { return self.onOptionHover.apply(self, arguments); });
-		$dropdown.on('mousedown click', '[data-selectable]', function() { return self.onOptionSelect.apply(self, arguments); });
-		watchChildEvent($control, 'mousedown', '*:not(input)', function() { return self.onItemSelect.apply(self, arguments); });
+		$dropdown.on('mouseup click', '[data-selectable]', function() { return self.onOptionSelect.apply(self, arguments); });
```
vanboom, line 195: Native selects change on the mouse-up.  Selectize should match this behavior.

src/selectize.js
```diff
@@ -231,7 +231,8 @@ $.extend(Selectize.prototype, {
 					return false;
 				}
 				// blur on click outside
-				if (!self.$control.has(e.target).length && e.target !== self.$control[0]) {
```
vanboom, line 234: $control will never have e.target here, as e.target is contained within $dropdown.  This appears to be a logic error.  When clicking inside of the dropdown we do not want to blur, so $dropdown.has(e.target) is the correct logic here.

src/selectize.js
```diff
@@ -231,7 +231,8 @@ $.extend(Selectize.prototype, {
...
+				// do not blur if the dropdown is clicked
+				if (!self.$dropdown.has(e.target).length && e.target !== self.$control[0]) {
```
vanboom, line 235: `e.target` is always contained within `self.$dropdown` here.  `$dropdown` and `$control` are sibling elements in the DOM.   `$dropdown` is the correct element to use for this logic.

src/selectize.js
```diff
@@ -648,12 +649,14 @@ $.extend(Selectize.prototype, {
 
 		if (self.ignoreFocus) {
 			return;
-		} else if (!self.ignoreBlur && document.activeElement === self.$dropdown_content[0]) {
-			// necessary to prevent IE closing the dropdown when the scrollbar is clicked
-			self.ignoreBlur = true;
-			self.onFocus(e);
-			return;
 		}
+		// Bug fix do not blur dropdown here
```
vanboom, line 653: This code causes the drop down to flash open then closed when a click is detected outside of the control.  Let's remove it - nobody uses IE11 any more.

risadams (approved):

PR #1831, opened 2022-07-09
Remove link for maintainer wanted

tnir:
New maintainer looks appointed as far as seen https://github.com/selectize/selectize.js/discussions/1678.

Signed-off-by: Takuya Noguchi <takninnovationresearch@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,6 @@
 
 ![Selectize.js](docs/selectize-wordmark.png)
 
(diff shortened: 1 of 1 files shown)


risadams:
not merging this,  but the link does point to the wrong place, so I'll get that updated :)

it's an open call for maintainers, so more help is always welcome.

tnir:
Thanks! e4116017ba75e497d3371f91039777fddba30e95

PR #1865, opened 2022-09-15
Fix #1818 Scroll left when I click on a selectized item and page is scrolled right

### LINKED ISSUE #1818: Scroll left when I click on a selectized item and page is scrolled right
I did:

	  like below

When I click on an option and if the page is scrolled right, there is a scroll left. 
With remove_button plugin, if I click on the cross, there is a scroll too and item is not removed.

Steps to reproduce:

1. go to https://selectize.dev/plugins/2020/01/01/remove-button-copy/
2. Add margin to example div like below and scroll to the right :

image

3. Try to remove one item

Expected result: Item is removed

Actual result: page scroll left and item is not removed

fabienwnklr:


src/selectize.js
```diff
@@ -984,7 +984,7 @@ $.extend(Selectize.prototype, {
 		var self = this;
 
 		self.setTextboxValue('');
-		self.$control_input.css({opacity: 0, position: 'absolute', left: self.rtl ? 10000 : -10000});
+		self.$control_input.css({opacity: 0, position: 'absolute', left: self.rtl ? 10000 : 0});
```
risadams, line 987: So if this is handling the position for RTL display, we probably want to change how this is done rather than pixel positioning. That can be a future taks.

risadams (approved):

PR #1872, opened 2022-09-21
Feature clear button

fabienwnklr:
New plugin `clear_button` displaying a button for clear input value(s)

Add two examples for `clear button` rendering (simple and multiple select)

Update doc for plugins (using SCSS instead LESS)

Don't hesitate to tel me any bad move or any improvement

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -13,7 +13,7 @@ complete control.
 - All plugins live in their own folders in ["src/plugins"](../src/plugins).
 - Plugin names should be in follow the format: `/[a-z_]+$`
 - JS source should live in a "plugin.js" file (required).
-- CSS should live in a "plugin.scss" file (optional) and include it into `src/scss/selectize.scss`. It will be bundled at build time.
+- CSS should live in a "plugin.less" and "plugin.scss" file (optional) if use SCSS file, include it into `src/scss/selectize.scss`. It will be bundled at build time.
 - Plugins are initialized right before the control is setup.
   This means that if you want to listen for events on any of the control's
   elements, you should override the `setup()` method (see ["DOM Events"](#dom-events)).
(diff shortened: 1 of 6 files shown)


src/plugins/clear_button/plugin.js
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+Selectize.define("clear_button", function (options) {
+  var self = this;
+
+  options = $.extend(
+    {
+      title: "Clear",
+      className: "clear",
+      libelle: "×",
```
risadams: should this be `Label`?

fabienwnklr: yes, sure , sorry keep the same doctrine

docs/plugins.md
```diff
@@ -13,7 +13,7 @@ complete control.
 - All plugins live in their own folders in ["src/plugins"](../src/plugins).
 - Plugin names should be in follow the format: `/[a-z_]+$`
 - JS source should live in a "plugin.js" file (required).
-- CSS should live in a "plugin.less" file (optional). It will be bundled at build time.
+- CSS should live in a "plugin.scss" file (optional) and include it into `src/scss/selectize.scss`. It will be bundled at build time.
```
risadams: For compatibility with the older versions of bootstrap, there should be both a sass and less version of styles.

This might change in the future if we decide to drop support for older versions, but there is enough of a user base that still does use older versions that we should try to support them where possible

risadams (changes requested):
Can you please also add a less stylesheet (it should mostly be a copy/paste) and check out the comments below?

Some unit tests would be great if time permits.

good job,  thanks!

docs/plugins.md
fabienwnklr: sure, misstake, I thought it had not been updated, I had trouble finding what I was missing to use my SCSS file

risadams (approved):

PR #1875, opened 2022-09-23
Feature : add 'size' option for custom dropdown height

fabienwnklr:
new 'size' option to set dropdown size, can be a number to displayed items, or a simple CSS height (px, rem, em, vh)

I also added two tests for value as number of items displayed and CSS value

Default value is `false` so nothing change

Open to improvement and patch

### DIFF
--- a/src/defaults.js
+++ b/src/defaults.js
@@ -54,7 +54,10 @@ Selectize.defaults = {
 	dropdownParent: null,
 
 	copyClassesToDropdown: true,
-  size: false, // must be a number items displayed or height (px, em, rem, vh)
+  dropdownSize: {
+    sizeType: 'auto', // 'numberItems' or 'fixedHeight'
+    sizeValue: 'auto', // number of items or height value (px is default) or CSS height (px, rem, em, vh)
+  },
   normalize: false,
 	/*
 	load                 : null, // function(query, callback) { ... }
--- a/src/defaults.js
+++ b/src/defaults.js
@@ -55,6 +55,7 @@ Selectize.defaults = {
 
 	copyClassesToDropdown: true,
   size: false, // must be a number items displayed or height (px, em, rem, vh)
+  normalize: false,
 	/*
 	load                 : null, // function(query, callback) { ... }
 	score                : null, // function(search) { ... }
--- a/src/defaults.js
+++ b/src/defaults.js
@@ -54,7 +54,7 @@ Selectize.defaults = {
 	dropdownParent: null,
 
 	copyClassesToDropdown: true,
-
+  size: false, // must be a number items displayed or height (px, em, rem, vh)
 	/*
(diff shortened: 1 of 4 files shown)


eduardo-mior:
Some time ago I needed this too.

I solved this with 1 line of CSS:

```css
.my-selectize .selectize-dropdown-content {
     max-height: none !important;
}
```
or
```css
.my-selectize .selectize-dropdown-content {
     max-height: 450px !important;
}
```

I particularly think that this kind of problem should be solved via CSS not JS.

fabienwnklr:
Yes for simple height it's easily doable with CSS,I implemented that CSS values be accepted for better customization and flexibility. But first improvement and first idea it's to set a count of visible value, like 5 items visible etc. 

Let's see what owner think about that !

src/defaults.js
```diff
@@ -54,7 +54,8 @@ Selectize.defaults = {
 	dropdownParent: null,
 
 	copyClassesToDropdown: true,
-
+  size: false, // must be a number items displayed or height (px, em, rem, vh)
```
risadams: I think I see what is going here, but it is a little confusing to overload size with multiple options.  It would be better to have two options (or perhaps an object?).  `sizeType`  that would be an enum of `auto` (default), `number of items`, `fixed height`.  and also a `sizeValue` that would take the number to use.

src/selectize.js
```diff
@@ -1091,7 +1091,8 @@ $.extend(Selectize.prototype, {
 		}
 
 		// perform search
-		if (query !== self.lastQuery) {
+    if (query !== self.lastQuery) {
+      if (settings.normalize) query = query.normalize("NFD").replace(/[\u0300-\u036f]/g, "");
```
risadams, line 1095: what does normalize do?   and what is the "NFD" value refer to?  are there any other options?

risadams (changes requested):
Got a couple of questions here.  Also Thanks for adding tests! :)

src/selectize.js
fabienwnklr: Sorry, I thought I sent this commit to another PR. it removes all accents in the string.
cf : [https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/String/normalize#param%C3%A8tres](url)

src/defaults.js
fabienwnklr: Hm i understand, I was inspired by another component that uses the 'size' thermal.

but an object like this seems like a good idea, if i understand correctly, it would look like this:
`
size: {
    sizeType: 'itemsNumber'
    sizeValue: 6
}
`
right?

src/selectize.js
risadams: Nice!  I learned something new today.

risadams (approved):

PR #1876, opened 2022-09-27
Fix #1776 deprecated $.fn.blur()

### LINKED ISSUE #1776: Close method calls deprecated `$.fn.blur()`
I did:

* [ ] Create a *minimal* working example on JsFiddle or Codepen
	  like below

The `close` method calls the deprecated `$.fn.blur()` shorthand:

```javascript
if (self.isBlurring) {
    self.$control_input.blur(); // close keyboard on iOS
}
```

This triggers a warning when running jQuery Migrate.

Steps to reproduce:

1. Click selectize element to open the items dropdown.
2. Click an item to select it. 
3. Click out of the selectize control

Expected result: No deprecation warning

Actual result: Deprecation warning

Further Details:

I think #1625 referenced this, but the PR which closed that issue (#1624) only actually addressed `$.fn.trim()`, not `$.fn.blur()`.

I wonder if this could be as simple as updating as follows, but I'm not confident in it:

```diff
if (self.isBlurring) {
-    self.$control_input.blur(); // close keyboard on iOS
+    self.$control_input[0].blur(); // close keyboard on iOS
}
```

Happy to create a PR if it is thought that this is the way to go! :-)

fabienwnklr:


risadams (approved):

mikevaux:
Thanks @fabienwnklr 😃

PR #1890, opened 2022-11-03
fix: position top / style for bootstrap 4/5

fabienwnklr:
Position top sometimes wrong, i fix it by replacing `prop('scrollHeight')`by `outerHeight(true)`

And i fix style box-shadow when dropdown is on top and with bootstrap 4/5 (not bootstrap 3 cause i don't know this version and it pretty outdated..)


risadams (approved):
It's fine that BS3 isn't updated, anything less than 4 is mostly legacy code at this point.

PR #1906, opened 2022-11-09
several fix and enhancements 

fabienwnklr:
- fix: style dropdown bootstrap4/5 : add `!important` for apply style when position is top with plugin `auto_position`
- fix: appareance with input-group class : border radius was incorrect if input into an input-group bootstrap
- feat: add ignoreOnDropdownHeight property : add property for ignore more selector for calculate height
- fix: calcul dropdownheight with optgroup : height was wrong when select has an optgroup
- fix: add separator height to dropdown height calcul : height was wrong when select has an optgroup
- feat: add option 'search': for having a single select appareance without search mode using input
- test: add test for normalize/search options : add test
- fix: persitant state style for selected item : keep style on selected option when we hover other option

> Note: sorry for the many commits I use two computers so sometimes I get lost...


src/selectize.js
```diff
@@ -158,6 +158,12 @@ $.extend(Selectize.prototype, {
 			$control_input.attr('placeholder', settings.placeholder);
 		}
 
+    // to have an identical rendering to a simple select (usefull for mobile device and do not open keyboard)
+    if (!self.settings.search) {
+      $control_input.attr('readonly', true);
```
eduardo-mior, line 163: There is a problem with ReadOnly.

If the person clicks on the input and types something on the physical keyboard the input will remain empty, but selectize will filter the items and will work anyway.

To fix this just listen to the focus event and when the input receives focus you apply a blur. That way the input will never be focused and the person will never be able to type anything and we won't have any problems.

eduardo-mior: A suggestion also to prevent the keyboard on mobile devices is to use $control_input.attr("inputmode", "none");

Some mobile browsers ignore the read_only property and show the keyboard anyway. But they respect input_mode.

The ideal is to have the 2 codes, read_only and input_mode none.

fabienwnklr: thank you so much for your feedback, I fixed that !

risadams (approved):

PR #1910, opened 2022-11-14
Missing reset for select element.

sdebacker:
Missing reset for select element that has a `form-select` class. Useful when using `selectize()` on a `<select>` element.


risadams (approved):

sdebacker:
Thanks for merging. Don’t forget to bump version.

PR #1924, opened 2022-11-30
Accessibility: measureString: hide temporary element from screen reader

meldafert:
The temporary element used by `measureString` is left in the document for future use.
However, while it is hidden visually, it will be visible to screen readers.
This PR adds the `aria-hidden` attribute which has no effect other than hiding it from screen readers.

Tested with NVDA 2022.3.2 and Firefox 107 on Windows 10.


risadams (approved):

risadams:
Good find!

PR #1927, opened 2022-12-01
fix(a11y): blur when tabbing out of input

### LINKED ISSUE #1926: Selectize element isFocused is true after $control_input gets blurred
I did:

	  like below

After [this commit that fixes an issue with modals](https://github.com/selectize/selectize.js/commit/e5792126a839095e00a743a5bade38caabeac4e5#diff-42f49af8bfaecef45ea594811f65dcb9aade0fbd8d3e7246d50ee92e863297caR210) , selectize component has stopped tracking $control_input blur events, so it keeps the "isFocused" state even when it is not focused anymore. That causes some visual errors as long as some other issues

#### Issue 1: Selectize input doesn't loose focus styles when tabbing through inputs.
Steps to reproduce:

1. Go to Demo page https://selectize.dev/docs/demos/lock
2. Click on Bootstrap 5, in order to get focused styles around the input field
3. Click on first select input
4. Press tab key to navigate to next input.

Expected result:
First input looses focus styles while second input shows them

Actual result:
Both inputs show focus styles.

#### Issue 2: When clicking on the dropdown of a multiple select, input loses focus and it doesn't gain it back because `onFocus` callback checks if isFocused is true before setting it ( which it is ), so you can't search using the input
 Steps to reproduce:

1. Go to Demo page https://selectize.dev/docs/demos/opt-groups
2. Type "B" in the input
3. Select "Bolts" in the dropdown using the mouse click
6. Type "C"

Expected result:
I expect the input to be focused so when I type the "C" it search for an option and shows it in the dropdown.

Actual result:
When I type "C" nothing happens because the input is not focused. When I click on the input to focus it it doesn't gain focus either, so the only way I can use the search again is clicking away from the drowdown menu (close it) and focusing the input again.

meldafert:
Fixes #1926.
To be accessible, it should be possible to easily navigate elements using the keyboard.
Previously, when using TAB to navigate through form elements, the input would enter its focused state when focused with TAB (and the dropdown would open when using `openOnFocus`).
However, when TAB is used to jump to the next focusable element, it would not lose its focused state, and the dropdown would stay open. This commit restores the behavior before #1813 and ensures that the input leaves its focused state and closes the dropdown when blurred using TAB.

### DIFF
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -218,8 +218,8 @@ $.extend(Selectize.prototype, {
 			keypress  : function() { return self.onKeyPress.apply(self, arguments); },
 			input     : function() { return self.onInput.apply(self, arguments); },
 			resize    : function() { self.positionDropdown.apply(self, []); },
-			// blur      : function() { return self.onBlur.apply(self, arguments); },
-			focus     : function() { self.ignoreBlur = false; return self.onFocus.apply(self, arguments); },
+			blur      : function() { return self.onBlur.apply(self, arguments); },
+			focus     : function() { return self.onFocus.apply(self, arguments); },
 			paste     : function() { return self.onPaste.apply(self, arguments); }
 		});
 
@@ -243,7 +243,12 @@ $.extend(Selectize.prototype, {
 				}
 				// blur on click outside
 				// do not blur if the dropdown is clicked
-				if (!self.$dropdown.has(e.target).length && e.target !== self.$control[0]) {
+				if (self.$dropdown.has(e.target).length) {
+					self.ignoreBlur = true;
+					window.setTimeout(function() {
+						self.ignoreBlur = false;
+					}, 0);
+				} else if (e.target !== self.$control[0]) {
 					self.blur(e.target);
 				}
 			}
@@ -685,19 +690,17 @@ $.extend(Selectize.prototype, {
 	 */
 	onBlur: function(e, dest) {
 		var self = this;
+
+		if (self.ignoreBlur) {
+			return;
+		}
+
 		if (!self.isFocused) return;
 		self.isFocused = false;
 
 		if (self.ignoreFocus) {
 			return;
 		}
-		// Bug fix do not blur dropdown here
-		// else if (!self.ignoreBlur && document.activeElement === self.$dropdown_content[0]) {
-		// 	// necessary to prevent IE closing the dropdown when the scrollbar is clicked
-		// 	self.ignoreBlur = true;
-		// 	self.onFocus(e);
-		// 	return;
-		// }
 
 		var deactivate = function() {
 			self.close();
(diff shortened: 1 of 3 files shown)


meldafert:
This breaks a test apparently. I will make sure to fix this, and to add a test case using TAB.

meldafert:
I have fixed the tests.
I also tried adding some tests for keyboard tabbing.
However, I was not able to get `syn.key()` to behave the way I want.
I emulated it using native `.focus()` for now, however that does not behave exactly the same way as actual keyboard tabbing, eg. for `selectOnTab`.
To get the best outcome, it would be necessary to use something like [puppeteer](https://www.npmjs.com/package/puppeteer-core) - this should be fairly easy as CI already has headless chrome set up.

src/selectize.js
```diff
@@ -218,8 +218,8 @@ $.extend(Selectize.prototype, {
 			keypress  : function() { return self.onKeyPress.apply(self, arguments); },
 			input     : function() { return self.onInput.apply(self, arguments); },
 			resize    : function() { self.positionDropdown.apply(self, []); },
-			// blur      : function() { return self.onBlur.apply(self, arguments); },
-			focus     : function() { self.ignoreBlur = false; return self.onFocus.apply(self, arguments); },
```
meldafert, line 222: `ignoreBlur` was not read anywhere anymore. I removed the previous references and re-used it for detecting when an item in the dropdown was clicked, to avoid closing the dropdown in that case.

test/api.js
```diff
@@ -633,7 +633,7 @@
 					test.selectize.search('hello');
 				}).to.not.throw(Error);
       });
-      it('should normalize a string', function () {
+      expect('should normalize a string', function () {
         var test;
 
         beforeEach(function() {
```
meldafert, line 639: This broke my new tests. Calling `beforeEach()` inside of `it()` (instead of inside `expect`) apparently applies it to all subsequent tests.

test/api.js
```diff
@@ -642,11 +642,12 @@
...
         });
 
-        it('should return query satinized', function() {
+        it('should return query satinized', function(done) {
           var query = test.selectize.search('héllo').query;
 
           window.setTimeout(function () {
             expect(query).to.be.equal('hello');
+            done();
           }, 0);
```
meldafert, line 651: Drive-by fix to ensure the test terminates properly.

test/interaction.js
```diff
@@ -413,6 +421,103 @@
...
+					expect(test.selectize.isOpen).to.be.equal(false);
+				});
+
+				// TODO: this would work if tabTo was using actual keyboard interactions,
+				// and not just focus()
+				xit('should select the first value on blur', async function() {
+					await tabTo(test.selectize.$control_input[0]);
+					await tabTo(input2[0]);
+					expect(test.selectize.getValue()).to.be.equal('a');
+				});
```
meldafert, line 481: This is the test that would require a better emulation/control over the headless browser (eg. puppeteer).

risadams:
Kudos to you.  this is one of the best documented pull requests I've seen in a long time :)
Thanks!

PR #1929, opened 2022-12-02
Plugin to make the selectize read-only

ggets:


fabienwnklr:
an option `search: true/false` already exists, why do this as a plugin ?

ggets:
I can't seem to find that option, please point me to it. This functionality is meant to be more like the lock(). The major benefit is that it is logically more compatible with html and css mainly because of the `readonly` attribute.

ggets:
The other benefit is that making it readonly doesn't allow user focus on the field which is rather confusing anyway on a input that is not allowed to be edited IMHO.

risadams:
I think it is a great plugin, I'm going to add some tests/examples and consider making it available in general

risadams (approved):

PR #1930, opened 2022-12-05
add custom rendergin example

fabienwnklr:


fabienwnklr:
I just saw the `Custom rendering and validation` example.. maybe we can split on 2 examples one `Custom rendering` and one `Validation` ? Because custom rendering can be quite wide-ranging, with `option` , `item`, `optgroup`, `optgroup_header` and `option_create` ?

src/selectize.js
```diff
@@ -1934,6 +1934,18 @@ $.extend(Selectize.prototype, {
...
+      values.forEach(function (item) {
+        var dataAttributes = self.$dropdown.find('.selected[data-value="' + item + '"]').data();
+        // Remove useless data attribute
+        if (dataAttributes) {
+          delete dataAttributes.value;
+          delete dataAttributes.selectable;
+        }
+
+        self.$input.find('option[value="' + item + '"]').data(dataAttributes);
+      })
```
risadams: I don't think that this should be part of this PR

risadams (changes requested):
We should probably not include static/js/* files in PR's,  the same way we exclude dist/* files.  It look like it will have the same problem with regeneration and conflicts.

The example by itself is fine,  but I think the data-attributes updates should be a separate feature once we have more time to devote to it.

PR #1935, opened 2022-12-12
add example, showArrow option and some fixes

fabienwnklr:
1. Add example using a custom width
2. add `showArrow` for removing arrow if necessary with simple CSS class and adapt style with it
3. Bind data attributes to orignal selected `<option>`
4. Fix input width when `fit-content` used

### DIFF
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -237,18 +237,15 @@ $.extend(Selectize.prototype, {
 
 		$document.on('mousedown' + eventNS, function(e) {
 			if (self.isFocused) {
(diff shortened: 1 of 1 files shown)


risadams (changes requested):
The demos are fine, but can we please split out the showArrow option (and also please add some documentation around it).

Let's also skip the data attributes for now until we have a better approach,  this seems to not be the right way

fabienwnklr:
> The demos are fine, but can we please split out the showArrow option (and also please add some documentation around it).
> 
> Let's also skip the data attributes for now until we have a better approach, this seems to not be the right way

done ! okay for you ?

PR #1936, opened 2022-12-12
Allow JSON as value select options

AlfonsoUceda:
Hello everyone:

I'm using selectize and I've found something that could be a "bug". In my select I use JSON like:
```
{"type":"place","name":"place_name for l 0","latitude":-27.7705,"longitude":-21.4252,"approxLocation":"city for l 0, province for l 0, country for l 0"}
```
for setting it as value in the option html tag.

The problem is that this line breaks because does not support it, it worked in v0.12.2.

Do you think this is ok?


risadams (approved):

AlfonsoUceda:
@risadams thanks you so much for your time ;)

PR #2028, opened 2023-08-17
Close #2027. Fix usage of `rgba()`

### LINKED ISSUE #2027: clear_button Sass code has issues compiling via LibSass
When trying to compile Sass -> Css via LibSass, I see:

```
Error: overloaded function `rgba` given wrong number of arguments
        on line 10:16 of inst/www/shared/selectize/plugins/clear_button/plugin.scss
        from line 80:9 of inst/www/shared/selectize/scss/selectize.scss
        from line 53:9 of inst/www/shared/selectize/scss/selectize.bootstrap5.scss
        from line 21:9 of stdin
>>         color: rgba(0, 0, 0);
```

Appears this problem could be solved by changing `rgba(0, 0, 0)` to `rgb(0, 0, 0)` (or just `black`):

cpsievert:


src/plugins/clear_button/plugin.scss
```diff
@@ -7,7 +7,7 @@
         width: 25px;
         top: 0;
         right: calc(#{$select-padding-x} - #{$select-padding-item-x});
-        color: rgba(0, 0, 0);
+        color: rgb(0, 0, 0);
```
cpsievert, line 10: This would work better with Bootstrap 5.3.1's new dark mode support, but I'm not sure if you want to incorporate Bootstrap semantics here

```suggestion
        color: var(--bs-emphasis-color, black);
```

databyte:
`rgba` is an alias of `rgb` so unless it's `rgb(0 0 0 / 50%)` it's not really a big deal. Merging since you already did the work!

PR #2059, opened 2023-10-20
Guarantee a unique event namespace

### LINKED ISSUE #2058: Can't add items to a (multi) select after updating inputs (in order)
[This codepen](https://codepen.io/cpsievert/pen/vYbBgEB?editors=1111) demonstrates the issue. In there, clicking the button (which updates the inputs with new options) breaks the 1st input's click-to-select behavior. Here's a screencast of the issue:

selectize-bug2

It's worth noting this isn't a problem with the latest release (`v0.15.2`), but it is a problem with the current  `HEAD` of the `master` branch (https://github.com/selectize/selectize.js/commit/e3f2e0b4aa251375bc21b5fcd8ca7d374a921f08), which is the version the codepen is above using.

Also, the codepen `console.log()`s some information related to (what appears to be) the source of the issue (`eventNS`). Here's my understanding of it.

1. Everytime `.selectize()` is called, it increases a global count, and uses that count for `eventNS`:

https://github.com/selectize/selectize.js/blob/e3f2e0b4aa251375bc21b5fcd8ca7d374a921f08/src/selectize.js#L29

2. In `.setup()` (which is called from `.selectize()`), that `eventNS` gets used to namespace events like this one:

https://github.com/selectize/selectize.js/blob/e3f2e0b4aa251375bc21b5fcd8ca7d374a921f08/src/selectize.js#L240-L240

3. Also,  the `.destroy()` method decrements that global count

https://github.com/selectize/selectize.js/blob/e3f2e0b4aa251375bc21b5fcd8ca7d374a921f08/src/selectize.js#L2379-L2382

As a result, if you initialize say 2 selectize inputs, the count will be at 2. If you then `.destroy()` the 1st widget (taking the count back to 1), then re-initialize it (bumping the count back to 2), then the the 1st widget winds up with the same `eventNS` as the 2nd widget.

Looking back at the history of this code gives us some insight into how this problem arose. Note that, [the `.destroy()` method was first added by @brianreavis, and introduced the `eventNS` concept](https://github.com/selectize/selectize.js/commit/1fbc3cda5ca15aa2c3879613b9701f15790ae1b5). In that first version, bumping the global count was fine, since the `.destroy()` method didn't decrement it. However, later on, [a decrement was added to destroy for another, unrelated reason](https://github.com/selectize/selectize.js/pull/1199).

It is a bit surprising to me that this issue has been around for nearly 7 years (and doesn't seem to be an issue for this particular behavior in `v0.15.2`), but I've been able to confirm, at least, that using a truly unique ID for `eventNS` will fix this particular problem (PR incoming)

cpsievert:
Closes #2058

### DIFF
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -29,7 +29,7 @@ var Selectize = function($input, settings) {
 		tagType          : input.tagName.toLowerCase() === 'select' ? TAG_SELECT : TAG_INPUT,
 		rtl              : /rtl/i.test(dir),
     // generate a unique ID for the event namespace
-		eventNS          : '.selectize-' + "-" + (new Date().getTime()) + "-" + (Math.random().toString(36).substr(2, 5)),
+		eventNS          : '.selectize-' + '-' + (new Date().getTime()) + '-' + (Math.random().toString(36).substr(2, 5)),
 
 		highlightedValue : null,
 		isBlurring       : false,
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -17,8 +17,8 @@ var Selectize = function($input, settings) {
 
   self.settings = {};
 
-  // increase the count of 'active' (i.e., non-destroyed) selectize instances
-  ++Selectize.count;
+	// increase the count of 'active' (i.e., non-destroyed) selectize instances
+	++Selectize.count;
 
 	// setup default state
 	$.extend(self, {
@@ -28,8 +28,8 @@ var Selectize = function($input, settings) {
 		tabIndex         : $input.attr('tabindex') || '',
 		tagType          : input.tagName.toLowerCase() === 'select' ? TAG_SELECT : TAG_INPUT,
 		rtl              : /rtl/i.test(dir),
-    // generate a unique ID for the event namespace
-		eventNS          : '.selectize-' + '-' + (new Date().getTime()) + '-' + (Math.random().toString(36).substr(2, 5)),
+		// generate a unique ID for the event namespace
+		eventNS          : '.selectize-' + (new Date().getTime()) + '-' + (Math.random().toString(36).substr(2, 5)),
 
 		highlightedValue : null,
 		isBlurring       : false,
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -19,6 +19,7 @@ var Selectize = function($input, settings) {
 
   // increase the count of 'active' (i.e., non-destroyed) selectize instances
   ++Selectize.count;
+
 	// setup default state
 	$.extend(self, {
 		order            : 0,
--- a/src/selectize.js
+++ b/src/selectize.js
@@ -17,6 +17,12 @@ var Selectize = function($input, settings) {
 
   self.settings = {};
 
+  // increase the count of 'active' (i.e., non-destroyed) selectize instances
+  ++Selectize.count;
+
+  // generate a unique ID for the event namespace
+  var eventNS = '.selectize-' + "-" + (new Date().getTime()) + "-" + (Math.random().toString(36).substr(2, 5));
+
 	// setup default state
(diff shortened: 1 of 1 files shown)


src/selectize.js
```diff
@@ -17,6 +17,12 @@ var Selectize = function($input, settings) {
 
   self.settings = {};
 
+  // increase the count of 'active' (i.e., non-destroyed) selectize instances
+  ++Selectize.count;
+
+  // generate a unique ID for the event namespace
+  var eventNS = '.selectize-' + "-" + (new Date().getTime()) + "-" + (Math.random().toString(36).substr(2, 5));
+
```
cpsievert:
```suggestion

```

src/selectize.js
```diff
@@ -25,8 +31,7 @@ var Selectize = function($input, settings) {
 		tabIndex         : $input.attr('tabindex') || '',
 		tagType          : input.tagName.toLowerCase() === 'select' ? TAG_SELECT : TAG_INPUT,
 		rtl              : /rtl/i.test(dir),
-
-		eventNS          : '.selectize' + (++Selectize.count),
+		eventNS          : eventNS,
```
cpsievert:
```suggestion
    // generate a unique ID for the event namespace
		eventNS          : '.selectize-' + "-" + (new Date().getTime()) + "-" + (Math.random().toString(36).substr(2, 5)),

```

src/selectize.js
```diff
@@ -17,6 +17,8 @@ var Selectize = function($input, settings) {
 
   self.settings = {};
 
+  // increase the count of 'active' (i.e., non-destroyed) selectize instances
+  ++Selectize.count;
```
cpsievert:
```suggestion
  // increase the count of 'active' (i.e., non-destroyed) selectize instances
  ++Selectize.count;

```

src/selectize.js
```diff
@@ -25,8 +28,9 @@ var Selectize = function($input, settings) {
 		tabIndex         : $input.attr('tabindex') || '',
 		tagType          : input.tagName.toLowerCase() === 'select' ? TAG_SELECT : TAG_INPUT,
 		rtl              : /rtl/i.test(dir),
+    // generate a unique ID for the event namespace
+		eventNS          : '.selectize-' + "-" + (new Date().getTime()) + "-" + (Math.random().toString(36).substr(2, 5)),
```
cpsievert:
```suggestion
		eventNS          : '.selectize-' + '-' + (new Date().getTime()) + '-' + (Math.random().toString(36).substr(2, 5)),
```

src/selectize.js
```diff
@@ -17,6 +17,9 @@ var Selectize = function($input, settings) {
 
   self.settings = {};
 
+  // increase the count of 'active' (i.e., non-destroyed) selectize instances
+  ++Selectize.count;
```
cpsievert:
```suggestion
	// increase the count of 'active' (i.e., non-destroyed) selectize instances
	++Selectize.count;
```

src/selectize.js
```diff
@@ -25,8 +28,9 @@ var Selectize = function($input, settings) {
 		tabIndex         : $input.attr('tabindex') || '',
 		tagType          : input.tagName.toLowerCase() === 'select' ? TAG_SELECT : TAG_INPUT,
 		rtl              : /rtl/i.test(dir),
+    // generate a unique ID for the event namespace
+		eventNS          : '.selectize-' + '-' + (new Date().getTime()) + '-' + (Math.random().toString(36).substr(2, 5)),
```
cpsievert:
```suggestion
		// generate a unique ID for the event namespace
		eventNS          : '.selectize-' + (new Date().getTime()) + '-' + (Math.random().toString(36).substr(2, 5)),
```

PR #2087, opened 2024-03-14
Right-pad dropdown text to prevent overlap with arrow

Mkranj:
When the text inside a dropdown is quite long, it is possible for it to obscure the arrow that indicates this is a dropdown menu. This can look somewhat unpolished.

I included a padding-right statement to preserve the bit of space that the arrow occupies.
Below is an example of a dropdown before and after this change.

selectize_rpad_arrow


databyte:
I can't quickly confirm this doesn't have any other side effects due to the broad selector used (` > * `). Also there's support for right-to-left languages such as Korean and Chinese that I'm sure doesn't assume the padding should always to be on the right.

PR #2109, opened 2024-06-11
Mak unreachable code inside setActiveItem with shift/ctrl mousedown

Yongjun042:
`setActiveItem` with event is called by `onItemSelect` and it is triggered by `watchChildEvent($control, 'mouseup', '*:not(input)', function() { return self.onItemSelect.apply(self, arguments); });`
So `eventName === 'mousedown'`is always false because eventName is mouseup.
Also ` (eventName === 'keydown' && this.isShiftDown)` is always flase too.


Yongjun042:
https://github.com/selectize/selectize.js/pull/1813/commits/e5792126a839095e00a743a5bade38caabeac4e5#r890628729
Changeing to mouseup was intended. Making a new pull request

